PR #61, opened 2026-02-03
Add files via upload

aifeifei798:
# OLD (Causing the crash)
if not self.model or not hasattr(self.model, 'tokenizer') or not hasattr(self.model, 'detokenizer'): # NEW (Fixed)
if self.model is None or not hasattr(self.model, 'tokenizer') or not hasattr(self.model, 'detokenizer'):


ChuxiJ:
already fixed

Issue #113, opened 2026-02-04
Gradio UI Hang/Freeze After Generations + Audio Not Loading (tested only in Windows) Solved


Quipjoy:
# 🐛Gradio UI Hang/Freeze After Second Generation + Audio Not Loading (Windows)  Solved both portable and install vers
## Testing Status & Acknowledgments
**Current Testing**: ✅ Verified on **portable version** (latest release from this morning, Feb 4th 2026)  
**Next Step**: 🔄 test on **installable version** (`git clone` + `uv sync`)  next msg
**Solution Found**: With the assistance of AI agents (Gemini & Sonnet) for systematic debugging and root cause analysis
---
## Problem Summary
Multiple users have reported critical UI issues with ACE-Step 1.5 on Windows:
1. **UI Controls Freeze** after the second generation (buttons unresponsive, requires hard refresh)
2. **Audio Files Not Loading** in Gradio players (files generated but not playable)
3. **Svelte Reconciliation Errors** in browser console: `TypeError: Cannot read properties of undefined (reading 'e')`
These issues are **100% reproducible** on Windows with Gradio 6.x and prevent normal usage of the application.
---
### 1. Audio Loading Failure
**Cause**: Path mismatch between `allowed_paths` configuration and actual audio file paths
- `allowed_paths` uses backslashes on Windows: `F:\repos\ACE-Step-1.5\gradio_outputs`
- Audio file paths use forward slashes: `F:/repos/ACE-Step-1.5/gradio_outputs/...`
- Gradio 6 rejects file serving due to path format mismatch
### 2. UI Freeze After Second Generation
**Cause**: Race conditions from multiple rapid progressive yields
- Progressive yields during generation trigger too many Gradio reactive updates
- Gradio's Svelte frontend cannot reconcile rapid state changes
- Results in UI deadlock requiring hard refresh
### 3. Svelte Reconciliation Error
**Cause**: Incorrect element order in yield tuples
- Frontend expects: Accordions (indices 30-37) → LRCs (indices 38-45)
- Backend was yielding: LRCs → Accordions (reversed order)
- Svelte reconciliation fails when element positions don't match
---
## Complete Solution
I've successfully tested and verified the following fixes on Windows (Chrome + Opera):
### 🔴 CRITICAL FIX 1: Normalize Path for Gradio 6 (acestep_v15_pipeline.py)
**This is the MANDATORY fix for audio loading on Windows:**
```python
[...]

PR #159, opened 2026-02-05
  fix: Windows path compatibility and UI freeze issues (Issue #113)

DumoeDss:
- Normalize paths with forward slashes for Gradio 6 compatibility on Windows
  - Add allowed_paths config to fix audio file serving
  - Add default_concurrency_limit=1 to prevent VRAM saturation
  - Disable progressive yields on Windows only to prevent UI freeze
  - Use local gradio_outputs directory instead of system temp
  - Fix path handling in batch navigation functions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -229,3 +229,4 @@ checkpoints_pack/
 issues/
 PortableGit/
 proxy_config.txt
+gradio_outputs/
--- a/acestep/acestep_v15_pipeline.py
+++ b/acestep/acestep_v15_pipeline.py
@@ -113,7 +113,15 @@ def main():
         print(f"CPU offload disabled by default (GPU >= 16GB)")
     else:
         print("No GPU detected, running on CPU")
-    
+
+    # Define local outputs directory
+    project_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    output_dir = os.path.join(project_root, "gradio_outputs")
+    # Normalize path to use forward slashes for Gradio 6 compatibility on Windows
+    output_dir = output_dir.replace("\\", "/")
+    os.makedirs(output_dir, exist_ok=True)
+    print(f"Output directory: {output_dir}")
+
     parser = argparse.ArgumentParser(description="Gradio Demo for ACE-Step V1.5")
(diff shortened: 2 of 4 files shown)


Quipjoy:
First off, thanks so much for making this repo and for merging the Windows path fixes—everything works great.

I noticed one small thing: 
in the installed version with the current implementation, audio files are being written to both locations:

✅ gradio_outputs/temp (local, as intended)
but too in Windows Temp folder
for me in C:\Users\...\AppData\Local\Temp\gradio\ 

I'm not sure if this duplication is intentional for you or just a side effect.
For sure it ends up duplicating the temp files.
Yesterday I local fixed for the portable version, but I haven't done the installed version yet, waiting for news/updates.
On my end, I patched it locally by setting the GRADIO_TEMP_DIR environment variable to redirect everything to the local folder. Here's what I added right after argument parsing in acestep_v15_pipeline.py:

```
python
# Configure Gradio to use local temp directory (prevent duplication in Windows temp)
output_dir = os.path.join(os.path.dirname(__file__), "..", "gradio_outputs")
os.makedirs(output_dir, exist_ok=True)
temp_dir = os.path.join(output_dir, "temp")
os.makedirs(temp_dir, exist_ok=True)
os.environ["GRADIO_TEMP_DIR"] = temp_dir
# Cleanup old temp files on startup
import shutil
if os.path.exists(temp_dir):
    try:
        shutil.rmtree(temp_dir)
        os.makedirs(temp_dir, exist_ok=True)
        print(f"[Cleanup] Cleared temp directory: {temp_dir}")
    except Exception as e:
        print(f"[Cleanup] Warning: Could not clear temp directory: {e}")
```

This way:

No more file duplication in Windows temp
Old temp files from previous sessions get cleaned up automatically on next server startup
Everything stays organized in the project folder

 just wanted to share in case anyone else runs into the same thing.

rkfg:
>if not IS_WINDOWS:

This issue is not windows-specific but rather universal. Something breaks inside Svelte's reconciler which makes the whole UI unresponsive, happens to me on Linux. I suggest deleting this entire code block. I did so in my fork and it hasn't happened to me since then. No negative effects either.

Issue #167, opened 2026-02-05
Quick support for cpu device?


TheephopWS:
Excited to see ACE-Step-1.5 released! I’m currently on a CPU-only setup and noticed that CPU training isn’t clearly documented, which apparently, the CPU training now isn't being supported.
Could you clarify whether CPU training is supported, or if it’s strictly inference-only? I hope this helps improve documentation and guide users without CUDA GPUs.

jayvenn21:
Short answer: ACE-Step supports **CPU inference**, but training (including LoRA) on CPU is not practical due to runtime constraints.

I’ve opened a small docs PR to clarify this and outline recommended alternatives (DiT-only mode, cloud GPUs, etc.) so this is easier to find going forward.

TheephopWS:
Thanks for the doc!

PR #170, opened 2026-02-05
Docs/cpu support clarification

jayvenn21:
This PR clarifies current CPU support to reduce confusion for users without CUDA GPUs.
	•	Documents that CPU is supported for inference only
	•	Explicitly notes that training on CPU is not practical
	•	Points users to DiT-only mode and alternative options
No functional changes.
for #167


ChuxiJ:
please solve the conflicts

README.md
```diff
@@ -75,6 +75,19 @@ Star ACE-Step on GitHub and be instantly notified of new releases
...
 > **Requirements:** Python 3.11, CUDA GPU recommended (works on CPU/MPS but slower)
 
+### CPU Support (Limited)
+
+ACE-Step can run on CPU for **inference only**, but performance will be significantly slower.
+
+- CPU inference is supported for testing and experimentation.
+- **Training (including LoRA) on CPU is not recommended** due to extremely long runtimes.
+- For low-VRAM systems, DiT-only mode (LLM disabled) is supported.
+- For training, a CUDA-capable GPU or cloud GPU instance is required.
```
Splendide-Imaginarius: Two lines above you say that CPU training is "not recommended", which is not the same thing as a GPU being "required" as this line says. Can you clarify whether you're simply saying that CPU training will be unpleasantly slow (better if you can provide an example of how slow we're talking... a week to train a LoRA? A year?), or that it literally does not work at all (e.g. due to code paths that only are implemented for GPU)?

jayvenn21: Good catch, thanks for pointing that out.

To clarify intent: CPU training may technically execute, but it is not practically usable due to extremely long runtimes and lack of optimization. There are no GPU-only hard failures in the codepath, but training on CPU is not an intended or supported workflow.

I’ll update the wording to reflect that CUDA is required *in practice* for any realistic training, rather than implying CPU training is impossible.

PR #176, opened 2026-02-05
feat: expose denoise control for audio-conditioned generation in Gradio

### LINKED ISSUE #133: audio-to-audio controlling the denoise parameter
**Is your feature request related to a problem? Please describe.**
I miss ACE-Step’s audio-to-audio workflow... where you could feed in a piece of music to resample and generate new material. Most importantly, you could control how different the result was from the original by adjusting the KSampler’s denoise parameter. Do I have to use ComfyUI ACE-Step workflows for v1.5? I’d prefer the Gradio interface.

**Describe the solution you'd like**
audio-to-audio workflow... cover doesn't seem to be the same as the original ace-step's approach to this...

**Describe alternatives you've considered**
Er... Using ComfyUI...

jayvenn21:
This PR exposes a denoise / similarity parameter in the Gradio UI when reference audio is used.

It restores the audio-to-audio resampling control present in earlier ACE-Step workflows,
without changing defaults or affecting text-only generation.

Fixes #133

### DIFF
--- a/acestep/gradio_ui/events/__init__.py
+++ b/acestep/gradio_ui/events/__init__.py
@@ -113,13 +113,19 @@ def setup_event_handlers(demo, dit_handler, llm_handler, dataset_handler, datase
     
     generation_section["init_llm_checkbox"].change(
         fn=gen_h.update_audio_cover_strength_visibility,
-        inputs=[generation_section["task_type"], generation_section["init_llm_checkbox"]],
+        inputs=[generation_section["task_type"], generation_section["init_llm_checkbox"], generation_section["reference_audio"]],
         outputs=[generation_section["audio_cover_strength"]]
     )
     
     generation_section["task_type"].change(
         fn=gen_h.update_audio_cover_strength_visibility,
-        inputs=[generation_section["task_type"], generation_section["init_llm_checkbox"]],
+        inputs=[generation_section["task_type"], generation_section["init_llm_checkbox"], generation_section["reference_audio"]],
+        outputs=[generation_section["audio_cover_strength"]]
+    )
+    
+    generation_section["reference_audio"].change(
+        fn=gen_h.update_audio_cover_strength_visibility,
+        inputs=[generation_section["task_type"], generation_section["init_llm_checkbox"], generation_section["reference_audio"]],
         outputs=[generation_section["audio_cover_strength"]]
     )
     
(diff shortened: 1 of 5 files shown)


Iris-Colt-IF:
Er... I downloaded the last version from the repository... and this commit doesn't seem to dynamically update the audio cover strength control's visibility and label when in "cover" mode when a reference audio is defined. What am I missing?

jayvenn21:
@Iris-Colt-IF  Thanks for checking this and flagging it 👍

The control is updated dynamically when reference audio is present in the audio-conditioned path, but the cover mode path doesn’t currently re-trigger the same UI update event. That’s likely what you’re running into.

This PR intentionally scoped the behavior to the audio-conditioned workflow to avoid changing cover defaults, but I agree the UX should be consistent. I’ll take a look at wiring the same visibility/label update into the cover-mode handler as a follow-up.

In the meantime, could you confirm:
	•	whether you’re running from main after the merge, and
	•	which mode you’re switching from → to when the reference audio is defined?

That’ll help me make sure the fix covers the right path.

Iris-Colt-IF:
> @Iris-Colt-IF Thanks for checking this and flagging it 👍

Thanks for your answer. I just downloaded the main after the merge. After inspecting the commit, I figured out that to expose the slider, I had to uncheck the "Initialize 5Hz LM" checkbox... then it appears under "Advanced Settings."

A couple of things, though: First, I'm not sure whether in the original ACE-Step workflow the audio you feed in is supposed to be the source audio or the reference audio. Second, I tried setting the slider to 1 (which, as I understand it, shouldn't affect the input music), but it's still modifying it. So I'm not sure if the current implementation matches the one in the original ACE-Step, or if I'm doing something wrong...

Iris-Colt-IF:
Wow, I figured it out! I need to provide a source audio as well. Now... I'm not sure what influence the reference audio has on the workflow, but as things stand, I can't just remove it to test whether source audio alone would suffice... because the slider disappears and I can no longer configure it. (That said, maybe it retains its value even when hidden.)

Iris-Colt-IF:
Okay, I've finally confirmed that the Reference Audio isn't taken into account at all (at least with my approach). My experiments indicate that only the Source Audio is used for audio-to-audio. For example, if I set the slider to any value (say, 1) and then remove the Reference Audio, the slider disappears but its effect persists (in this case, doing nothing). This more or less confirms that the workflow has nothing to do with Reference Audio. Am I right or the Reference Audio has still a role in audio-to-audio?

PR #194, opened 2026-02-05
Feat/progress UI: Prevent Gradio slider math error when no audio files are found

1larity:
Adds a safe slider helper to ensure a non-zero range, avoiding math domain error.
Uses the helper in scan/load paths when the dataset is empty or missing.
Hides the slider when there are zero or one samples to keep the UI clean.

Fix text encoder device alignment (see review below)

Fix dataset JSON not loading to UI correctly.


1larity:
Added a fallback so the Lyrics textbox shows formatted_lyrics when lyrics is empty.

When loading a dataset, it now updates dataset‑level UI controls (name, custom tag, tag position, 
instrumental, genre ratio) from the loaded metadata.

Added a load status note when formatted lyrics are being shown instead of lyrics.

1larity:
### Functional Decomposition of dataset_builder.py complete. 22 new files from monolithic source file, placed in _\acestep\training\dataset_builder_modules_.

**Minor fixes:-**
Lyrics objects intent adherence
When a user edits Lyrics (used for training), we now also update formatted_lyrics to the same value (unless instrumental, then it clears).

This aligns intent: lyrics = training truth, formatted_lyrics = processed/user‑approved text.

UI refresh fix with single sample
Auto‑label now forces UI refresh of the preview/edit panel after completion, so with 1 sample the preview isn’t left blank.
Status text appends “✅ Preview refreshed.” so you can see the refresh chain ran.

Device mismatch fix in text encoding
 (mask)
text_attention_mask is moved to the same device as the text encoder when text_input_ids is moved, preventing device mismatch in model.encoder.

**Copilot review, only important observations:**

_Edge Case: Empty Dataset:_
If no samples are available in the dataset, the get_sample_preview function might not handle this gracefully. Ensure it can handle empty or invalid sample_selector inputs.
Performance:
Refreshing all fields (e.g., edit_caption, edit_lyrics, etc.) for every sample might be inefficient for large datasets. Consider optimizing by refreshing only the fields that changed.
Error Handling:
The lambda function assumes status and has_raw inputs are valid. If these are None or malformed, it could lead to runtime errors.

_Edge Case: Empty or Null Lyrics:_
If lyrics is None or an empty string, updated_formatted will also be empty. Ensure downstream processes can handle this.

_Regression Risk:_
If other parts of the code rely on formatted_lyrics being non-empty, this change might introduce unexpected behavior.

_Edge Case: Device Mismatch:_
If text_encoder parameters are split across multiple devices (e.g., model parallelism), the .to(text_dev) logic might fail. Ensure this handles such scenarios.
Performance:
Moving text_input_ids and text_attention_mask to the device twice (once for device and once for text_dev) could introduce unnecessary overhead. Consider consolidating these operations.

**Review/rebuttal (scoped to my changes only)**

Empty dataset handling: addressed. get_sample_preview now guards None selector and empty sample list.
_Refresh efficiency:_ the refresh is only triggered after auto‑label completes, not per sample; acceptable for UI correctness. No per‑row re-render loop added.
_None checks:_ status message now defaults when label_progress is false; has_raw is cast to bool to avoid malformed inputs.
_Empty/null lyrics:_ intended behaviour is to allow empty formatted_lyrics when user clears lyrics; training uses lyrics. No downstream failures observed.
formatted_lyrics non‑empty assumption: updated to reflect user intent; this was a deliberate behavior change.
_Device mismatch:_ fixed in text encoding by moving attention mask with input ids; model‑parallel edge case is theoretical and out of scope for this commit.

Net: changes are intentional, scoped, and improve UI stability without altering core training behaviour.

ChuxiJ (approved):

PR #68, opened 2026-02-04
Add MPS support options and fix Gradio batch skip

pincente:
## Add `mps` as a device option in CLI and UI.
Improve Apple Silicon GPU detection with an MPS VRAM heuristic and override.
- Avoid Gradio FileData validation errors by using `gr.skip()` in batch nav fallbacks.\n\n## Testing
- Manual: Gradio UI on Apple Silicon with MPS device.
- Manual: Generated audio and batch navigation without FileData errors.

## Notes
- MPS memory estimate is heuristic; override via `MAX_MPS_VRAM`.

PR #247, opened 2026-02-06
gpu_config for MPS support

chigkim:
Claude implemented Apple MPS GPU detection and tier configuration in acestep/gpu_config.py so all entry points (CLI, API server, Gradio UI) get MPS tiering automatically.

It fixes the problem where it reports 0GB memory and assigns tier 1 on MPS regardless memory capacity.

MPS uses unified memory, but flash attention is not available. Thus, it consumes much more memory than dedicated CUDA VRAM.

Right now the tier assignments are just guesstimates, so it would need adjustment.

### DIFF
--- a/acestep/gpu_config.py
+++ b/acestep/gpu_config.py
@@ -2,16 +2,23 @@
 GPU Configuration Module
 Centralized GPU memory detection and adaptive configuration management
 
+Supports CUDA, Intel XPU, and Apple MPS (Metal Performance Shaders) backends.
+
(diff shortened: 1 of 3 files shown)


ChuxiJ:
duplicate: 
- https://github.com/ace-step/ACE-Step-1.5/pull/80
- https://github.com/ace-step/ACE-Step-1.5/pull/68

Don’t just submit PRs without resolving the conflicts.

chigkim:
This pr doesn't  have any conflict with main branch. That said, feel free to close since there's already another related pr.

PR #265, opened 2026-02-07
Add tokenizer cache, and offloading for acestep-5Hz-lm-4B to fit on 8GB vram 

nalexand:


nalexand:
Tested on 8 Gb VRAM all models 0.6b 1.7b 4b, quality not very good, problem with loudness (sometimes to low but most time to loud over limit), and sound not clean, not production quality. [HeartMula](https://youtu.be/SVE6idyNpx0) do better quality but less controlable, tags not work.

ChuxiJ:
If you aren't getting better results than HeartMula, you must be using it incorrectly, or there’s an environment issue causing bugs in generation. HeartMula is genuinely terrible.

PR #295, opened 2026-02-07
feat(MLX) MLX Native Inference Acceleration

tonyjohnvan:
## Summary

This PR adds a native MLX backend for the 5Hz LM inference engine, providing significant performance gains on Apple Silicon Macs by running transformer forward passes directly on the Metal GPU through Apple's unified memory architecture, instead of going through PyTorch's MPS translation layer.

- Add complete MLX inference backend with auto-detection, graceful fallback, progress reporting, and CFG support
- Fix a silent failure bug where `generate_from_formatted_prompt()` rejected the MLX backend due to a `self.llm is None` guard
- Add `mlx` as a selectable backend across CLI, REST API, and Gradio UI
- Add platform-gated `mlx` / `mlx-lm` dependencies (Apple Silicon only, zero impact on other platforms)

### 1. `acestep/llm_inference.py` — Core MLX Backend Implementation

This file contains the bulk of the changes (~440 lines added across 3 commits). Every change is detailed below:

#### a) New instance variables in `__init__()` (lines 57–59)

```python
self._mlx_model = None
self._mlx_model_path = None
```

Added to hold the MLX model instance and its filesystem path. These are separate from `self.llm` (which holds the PyTorch/vllm model) because MLX models are a fundamentally different object type — they are `mlx.nn.Module` instances backed by Metal GPU memory, not `torch.nn.Module`. The `_mlx_model_path` is stored separately because the MLX model object doesn't carry its original path, and we need it later to load a HuggingFace PyTorch model for quality scoring.

#### b) Auto-detection logic in `initialize_5hz_lm()` (lines 411–437)

When `backend == "mlx"` is explicitly selected, **or** when `backend == "vllm"` but the device is `"mps"` (Apple Silicon), the code now auto-detects MLX availability and prefers it. This is the key UX decision: on a Mac, vllm/CUDA doesn't work anyway, so we automatically try the best-performing backend first. The fallback chain is:

- **Explicit `"mlx"`**: Try MLX → fall back to PyTorch with warning
- **`"vllm"` on MPS**: Try MLX → fall through to vllm attempt → see (c) below
- **MLX not installed**: Log warning, fall back to PyTorch

This ensures users on Apple Silicon get optimal performance without changing any configuration.

#### c) Improved vllm failure fallback (lines 443–461)

When vllm initialization fails on MPS devices, the code now tries MLX *before* falling back to plain PyTorch. The complete fallback order becomes: **vllm → MLX → PyTorch**. The `else:` branch was changed to `elif backend != "mlx":` to prevent the PyTorch branch from running when the MLX case was already handled above — if `backend` is `"mlx"`, the function has already returned by this point.

#### d) `_is_mlx_available()` static method (lines 2470–2478)

A simple availability check that attempts to import both `mlx.core` and `mlx_lm`. Returns `False` on any `ImportError`. This avoids crashes on non-Apple-Silicon systems and is called before any MLX code path is entered.

#### e) `_load_mlx_model()` (lines 2480–2522)

Loads the model using `mlx_lm.utils.load()`. This function reads HuggingFace safetensors files directly — no weight conversion step is needed. `mlx-lm` uses the model's `config.json` to determine the architecture (Qwen3 in ACE-Step's case) and its `sanitize()` method handles the weight key remapping between HuggingFace and MLX naming conventions. The `mx.eval()` call materializes all parameters into Apple Silicon's unified memory. On success, sets `self.llm_backend = "mlx"` and `self.llm_initialized = True`.

#### f) `_make_mlx_cache()` (lines 2524–2537)

Creates KV caches for autoregressive generation. Tries the `mlx_lm.models.cache.make_prompt_cache` API first (the standard approach in mlx-lm >= 0.20.0), then falls back to the model's own `make_cache()` method for compatibility with different mlx-lm versions.

#### g) `_run_mlx_single()` (lines 2539–2749)

The core single-item generation method. Uses a **hybrid MLX/PyTorch architecture**:

- **MLX handles the forward pass** — matrix multiplications run natively on Apple's Metal GPU via unified memory, which is where the performance gain comes from (no MPS translation layer overhead, no host↔device copies)
- **PyTorch handles logits processing and sampling** — constrained decoding, top-k/p filtering, repetition penalty, and token sampling all reuse the existing tested code paths

The bridge between frameworks works by: `MLX bfloat16 tensor → mx.astype(float32) → numpy array → torch.from_numpy()`. The float32 cast is necessary because numpy doesn't support bfloat16.

**CFG (Classifier-Free Guidance) support**: When `cfg_scale > 1.0`, the method maintains two separate KV caches (conditional and unconditional), runs both forward passes in MLX, applies the CFG formula `uncond + scale * (cond - uncond)` in MLX tensor space, then bridges the combined logits to PyTorch for sampling.

**Progress reporting**: The prefill phase is timed and logged with throughput (`tok/s`). The autoregressive loop is wrapped with `tqdm(total=max_new_tokens, desc="MLX [CFG ]Generation", unit="tok")` matching the PyTorch backend's progress bar style. A summary line is logged after generation completes showing total tokens, decode throughput, and a prefill+decode time breakdown.

#### h) `_run_mlx()` (lines 2751–2836)

Unified entry point supporting both single and batch modes. Batch items are processed sequentially (same approach as the PyTorch backend). Sets MLX random seeds per batch item for reproducibility.

#### i) Codes generation dispatch (lines 1153–1170)

Added `elif self.llm_backend == "mlx":` in the Phase 2 batch codes generation path so the "codes" phase (generating audio tokens) uses `_run_mlx()` instead of falling through to the `_run_pt()` default.

#### j) CoT generation dispatch (lines 2051–2074)

Added `elif self.llm_backend == "mlx":` in `generate_from_formatted_prompt()` so the Chain-of-Thought reasoning phase and single-item generation also route to the MLX backend.

#### k) `generate_from_formatted_prompt()` null-check fix (lines 2002–2008)

**Bug fix**: The original guard `if self.llm is None or self.llm_tokenizer is None` silently returned an empty string when the MLX backend was active, because MLX sets `self._mlx_model` but leaves `self.llm` as `None`. This killed:

- **Phase 1 (CoT metadata generation)** — returned empty, so no caption, language, BPM, key, or time signature were generated
- **Single-mode Phase 2** (line 1238) — also routes through this function

The bug was masked in the default Gradio UI because `batch_size=2` with pre-filled metadata skips Phase 1 entirely ("Using user-provided metadata") and Phase 2 batch mode calls `_run_mlx()` directly, bypassing `generate_from_formatted_prompt()`.

**Fix**: The check now dispatches on `self.llm_backend`: if `"mlx"`, checks `self._mlx_model`; otherwise checks `self.llm`. Other `self.llm is None` checks in the file (lines 2153, 2253, 2862) are safe — they live inside PyTorch-only methods (`_generate_with_constrained_decoding`, `_generate_with_cfg_custom`) or in `_load_model_context` which already skips for the MLX backend.

#### l) Model offloading guard (line 2853)

Added `"mlx"` to the set of backends that skip CPU offloading in `_load_model_context()`. MLX manages its own memory on Apple Silicon's unified memory — there is no discrete GPU to offload from, so the offload cycle would be both unnecessary and harmful.

#### m) Scoring model support in `get_hf_model_for_scoring()` (lines 2932–2961)

Added `elif self.llm_backend == "mlx":` handler. MLX model objects can't be used with PyTorch's `CrossEntropyLoss` for quality scoring, so a separate HuggingFace `AutoModelForCausalLM` is loaded on-demand from the stored `_mlx_model_path` — the same pattern used by the vllm backend. The model is placed on MPS (or CPU as fallback) since scoring is not performance-critical.

---

### 2. `acestep/acestep_v15_pipeline.py` — CLI Argument

**One line changed**: Added `"mlx"` to the `choices` list for the `--backend` argument and updated the help text. This lets users explicitly select the MLX backend from the command line with `--backend mlx`.

---

### 3. `acestep/api_server.py` — REST API Backend Validation (4 locations)

Every place that validates the `backend`/`lm_backend` parameter now includes `"mlx"`:

| Location | What |
|---|---|
| Line 513 | Pydantic `Literal["vllm", "pt", "mlx"]` type hint in `GenerateMusicRequest` |
| Line 1204 | Backend validation in on-demand LLM initialization path |
| Line 1890 | Backend validation during server startup LLM initialization |
| Line 2372 | Backend validation in auto-labeling endpoint |

All four are superset expansions — existing `"vllm"` and `"pt"` values are accepted identically. Without these changes, the API server would silently reset `"mlx"` to `"vllm"` in the validation fallback.

---

### 4. `acestep/gradio_ui/interfaces/generation.py` — Gradio UI Dropdown

**One line changed**: Added `"mlx"` to the `choices` list of the backend dropdown in the service initialization panel. Users can now select "mlx" from the web interface.

---

### 5. `pyproject.toml` — Package Dependencies

**Two lines added** with platform environment markers:

```
"mlx>=0.25.2; sys_platform == 'darwin' and platform_machine == 'arm64'"
"mlx-lm>=0.20.0; sys_platform == 'darwin' and platform_machine == 'arm64'"
```

The markers ensure these packages are **only installed on Apple Silicon Macs** (`darwin` + `arm64`). They are never installed on Linux, Windows, or Intel Macs. `mlx>=0.25.2` provides the core Metal-accelerated array framework; `mlx-lm>=0.20.0` provides model loading utilities and KV cache management for transformer architectures.

---

### 6. `requirements.txt` — Pip Dependencies

The same two lines with identical platform markers, plus descriptive comments. This mirrors the `pyproject.toml` change for users who install via `pip install -r requirements.txt` instead of `uv sync`.

---

## Regression Analysis

All files pass syntax checking. Here is the complete regression analysis:

### Category 1: Purely additive code (ZERO risk)

- New MLX methods in `llm_inference.py` (`_is_mlx_available`, `_load_mlx_model`, `_make_mlx_cache`, `_run_mlx_single`, `_run_mlx`) — never called by existing vllm/pt code paths
- New `elif self.llm_backend == "mlx":` branches in 3 dispatch points — only entered for the new backend
- New `self._mlx_model` and `self._mlx_model_path` attributes in `__init__` — never read by existing code
- New `elif self.llm_backend == "mlx":` in `get_hf_model_for_scoring()` — only entered for new backend

### Category 2: Expanded allowed values (ZERO risk)

- `choices=["vllm", "pt"]` → `choices=["vllm", "pt", "mlx"]` in 3 files (CLI, Gradio, API)
- `{"vllm", "pt"}` → `{"vllm", "pt", "mlx"}` in 3 validation checks in API server
- Superset of original values. Existing `"vllm"` and `"pt"` still accepted identically.

### Category 3: Conditional dependencies (ZERO risk on non-macOS)

- `mlx>=0.25.2` and `mlx-lm>=0.20.0` gated by `sys_platform == 'darwin' and platform_machine == 'arm64'`
- Never installed on Windows or Linux

### Category 4: Modified control flow

| Change | On CUDA/Linux/Windows | On Apple Silicon (MPS) |
|---|---|---|
| `initialize()` MLX auto-detect block | `device != "mps"`, block skipped, **SAME behavior** | Tries MLX first. If fails, falls through to original vllm→PT path. **SAME or better** |
| `initialize()` vllm failure handler | `device != "mps"`, new MLX block skipped, **SAME behavior** | Tries MLX before PT fallback. **SAME or better** |
| `else:` → `elif backend != "mlx":` | `backend` is `"vllm"` or `"pt"`, condition true, **SAME behavior** | If `"mlx"`, already returned above. If `"pt"`, condition true, **SAME behavior** |
| `generate_from_formatted_prompt()` null check | `llm_backend` is `"vllm"`/`"pt"`, enters `elif`, checks `self.llm`. **SAME behavior** | If `"mlx"`, checks `_mlx_model`. If `"pt"`, checks `self.llm`. **SAME behavior** |
| `_load_model_context()` expanded skip set | `"vllm"` still in set, **SAME behavior**. `"pt"` still not in set, **SAME behavior** | `"mlx"` now in set (skips offload). Correct for unified memory. |

**Conclusion: NO REGRESSIONS.** Every existing code path on every platform produces identical behavior. The only observable difference is on Apple Silicon where the LLM now runs on native MLX instead of PyTorch MPS — a pure performance improvement.

## Test Plan

- [ ] Verify `uv sync` / `pip install` on Linux/Windows does NOT install `mlx` or `mlx-lm`
- [ ] Verify existing `--backend vllm` and `--backend pt` work identically on CUDA
- [ ] Verify `--backend mlx` on Apple Silicon: model loads, Phase 1 CoT generates metadata, Phase 2 generates audio codes
- [ ] Verify `batch_size=1` works (exercises `generate_from_formatted_prompt()` null-check fix)
- [ ] Verify `batch_size=2` works (exercises `_run_mlx()` batch path)
- [ ] Verify tqdm progress bar and tok/s appear during MLX generation
- [ ] Verify auto-detection: `--backend vllm` on MPS auto-selects MLX
- [ ] Verify graceful fallback: if `mlx-lm` is uninstalled, falls back to PyTorch

### DIFF
--- a/acestep/llm_inference.py
+++ b/acestep/llm_inference.py
@@ -2002,7 +2002,11 @@ class LLMHandler:
         """
         if not getattr(self, "llm_initialized", False):
             return "", "❌ 5Hz LM not initialized. Please initialize it first."
-        if self.llm is None or self.llm_tokenizer is None:
+        # Check that the appropriate model is loaded for the active backend
+        if self.llm_backend == "mlx":
+            if self._mlx_model is None or self.llm_tokenizer is None:
+                return "", "❌ 5Hz LM is missing MLX model or tokenizer."
+        elif self.llm is None or self.llm_tokenizer is None:
             return "", "❌ 5Hz LM is missing model or tokenizer."
 
         cfg = cfg or {}


tonyjohnvan:
@ChuxiJ Please take a look at your convince, this branch initiate an early stage Native MLX support with integration of the MLX-Audio https://github.com/Blaizzy/mlx-audio, tested on M3Max MBP with roughly 2x-7x performance gain on CFG Generations. 

| Before | After |
| :---: | :---: |
| before | after |

ChuxiJ (approved):

ChuxiJ:
For LM, use mlx-vllm whenever possible; for DiT, use pure MLX. Thanks for your contribution—we still hope to find a faster solution for LM.

Once it's fully tested, let me know and I'll merge the PR.

ChuxiJ:
https://github.com/waybarrios/vllm-mlx

PR #303, opened 2026-02-07
fix(training): remove invalid output_dir debug in auto_label_all

1larity:
Fixes a regression in Dataset Builder where Auto-Label All crashed with:

NameError: name 'output_dir' is not defined

What changed
Removed an invalid debug log line in auto_label_all that referenced output_dir, which is not in that function’s scope.
No functional behaviour changes to labelling logic beyond preventing the crash.

Why
Auto-Label All does not take an output_dir input in the UI or handler signature.
The debug line was mistakenly copied from preprocessing context and caused runtime failure before labelling started.
User impact
Auto-Label All now runs again instead of failing immediately.
No UI changes required.

The broken auto_label_all() reference was the only direct occurrence of this exact issue.
Remaining output_dir uses are in functions that actually take output_dir as a parameter 
(preprocess_dataset, preprocessing module), so they are valid.
No other obvious [debug_log_for(...)] call in training handlers is referencing missing locals.
So: no additional matches of this specific unwanted behaviour found.

_______________________________
Copilot review (TLDR no issues)

Analysis:
Crash Fix Verification:

The removed line references output_dir, which is not defined in the function. This would cause a NameError: output_dir is not defined when the function is executed.
Removing this line resolves the crash, as the undefined variable is no longer accessed.
Regression Risk:

The removed line is a debug log statement and does not affect the core functionality of auto_label_all.
No other references to output_dir exist in the function, so its removal does not introduce any regression risk.
Functional Changes:

The only change in this commit is the removal of the debug log line. No other functional changes are introduced.
Conclusion:
The commit correctly fixes the NameError crash.
There is no regression risk in the auto_label_all behavior from removing this debug line.
No functional changes beyond the crash fix are present in this commit.

### DIFF
--- a/acestep/gradio_ui/events/training_handlers.py
+++ b/acestep/gradio_ui/events/training_handlers.py
@@ -110,8 +110,6 @@ def auto_label_all(
     if llm_handler is None or not llm_handler.llm_initialized:
         return builder_state.get_samples_dataframe_data(), "� LLM not initialized. Please initialize the service with LLM enabled.", builder_state
 
-    debug_log_for("dataset", f"UI preprocess_dataset: output_dir='{output_dir.strip()}'")
-
     def progress_callback(msg):
         if progress:
             try:


1larity:
Beat me to it, @ChuxiJ, disregard. 

Many apologies for inconvenience, not sure how Codex, Copilot and Devstral all missed this.

Issue #308, opened 2026-02-07
Expected all tensors to be on the same device


CognitiveDiffusion:
Now Tensor preprocessing is acting up. It's still running at ~700 seconds for 37 songs, but I got this error in the console:

`RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument index in method wrapper_CUDA__index_select)`

There was no way to choose only one device and for initialization "Auto" was recommended.

1larity:
Fix in #309

Issue #292, opened 2026-02-07
Tensor Preprocess Crash: FlashAttention Invoked on CPU in run_encoder After Very Slow CPU BF16 VAE Encode


1larity:
Overall Analysis
with my new debugger code I  got past the a VAE-encode crash when Preprocessing tensors on low VRAM GPU, but hit a new pipeline/config mismatch during run_encoder.

Timeline from debug log:

Model init chooses flash_attention_2.
Pre-process starts.
VAE runs on CPU in bfloat16 and takes 6,443,544 ms (~1h47m) for a 58s clip.
Text/lyrics encode complete.
run_encoder crashes with:
NotImplementedError: flash_attn::_flash_attn_varlen_forward ... CPU backend

**The failure is deterministic and is not random instability.**

Detailed Root-Cause
This is a multi-factor config/path issue:

Flash attention is active while execution is on CPU path.
Error proves flash_attn kernel was dispatched on CPU.
flash_attn op is CUDA-only in Ace-step stack.
dtype got pushed to bfloat16 in a CPU path.
Log shows: vae device=cpu vae dtype=torch.bfloat16.
**That explains the huge VAE time (CPU bf16 path is very slow on Windows desktop - CPU).**

Training preprocess path calls encoder directly instead of via model-load context.
In normal inference, handler has load/offload context management.
Preprocess bypassing that means model/device/attention backend mismatch is exposed directly.
Most likely trigger sequence:
Flash attention enabled at init.
Model/vae in CPU/offload path (or CPU-selected path).
**Preprocess hits encoder with flash-attn still configured -> hard failure.**

@Kwisss

Will this be a problem on 16GB GPU?

If 16GB run is true CUDA, no CPU offload, and model stays on GPU: likely not this exact crash.
If 16GB run still uses CPU/offload path for preprocess (or auto selects CPU): yes, same class of failure can happen.
Even without crash, CPU bf16 VAE encode will still be painfully slow.
So 16GB VRAM is sufficient to avoid this if device/load settings are correct, but not sufficient by itself if config routes preprocess to CPU with flash-attn enabled.

Proposed fixes, priority order.

For preprocessing runs, force device=cuda (not auto) and potentially disable CPU offload*.
Disable flash attention whenever model/encoder may run on CPU.
Ensure preprocess encoder path uses same device/offload guard as inference path.
Prevent CPU+flash-attn combination at init (hard guard).
[...]

PR #309, opened 2026-02-07
Tensor preprocessing & Lora fix

1larity:
## PR: Improve Training Reliability on Low-VRAM Systems

Resolves multiple OOM, device residency, feature extraction issues.
#292 #308 #167 
---

This PR improves **training reliability on low-VRAM systems** by:

- Automatically switching **quantized inference presets** to a **training-safe, non-quantized configuration**

- Hardening **CPU/GPU tensor residency and dtype handling** to prevent device mismatch crashes during preprocessing and LoRA training

-Feature added, when training debug flag is set DEBUG_TRAINING = "VERBOSE", Epoch may be set below 100 for quick test cycles.

---

### **LoRA training could run against quantized DiT and fail**
- **Bug:** LoRA injection (PEFT) is incompatible with quantized decoder modules in this runtime.
- **Fix:** Added quantization guardrails and a best-effort auto-switch to a **training preset with quantization disabled**, with clear user-facing status and error messages.

---

### **Training flow could proceed with invalid training state input**
- **Bug:** Training state could be non-dict and break downstream assumptions.
- **Fix:** Added a **state normalization fallback** in the training event wrapper.

---

### **TensorBoard logger initialization could hard-fail training startup**
- **Bug:** Missing TensorBoard modules could cause logger setup to abort training.
- **Fix:** Logger creation is now **optional**; training proceeds without TensorBoard if unavailable.

---

### **Preprocessing encoder path could crash from CPU/CUDA mismatch**
- **Bug:** Encoder inputs could reside on a different device or dtype than the active DiT model, causing backend or flash-attention failures.
- **Fix:** Run the encoder under the **model load context** and explicitly align all relevant tensors to the model’s device and dtype before execution.

---

### **Lyrics encoding could fail due to device mismatch**
- **Bug:** Token tensors could be created on a device different from the text encoder parameters, causing `embedding` / `index_select` failures.
- **Fix:** Explicitly align lyric input tensors to the **actual text encoder device**.

---

### **Training UI could report success after internal failure**
- **Bug:** Certain failure states were treated as normal progress or completion.
- **Fix:** Added **failure detection** in the progress loop and an explicit **failed finalization path**.

---

### **VAE dtype/device handling unstable for low-VRAM offload paths**
- **Bug:** CPU-offloaded VAE dtype handling was not tier-aware and could be slow or fragile.
- **Fix:** Added **tier-aware CPU dtype selection** (e.g. `float32` on lower-VRAM tiers) and explicit VAE offload dtype handling.

---

### **No reusable initialization context for preset switching**
- **Bug:** Training-safe reinitialization could not reliably occur without access to prior init parameters.
- **Fix:** Persist the **last successful initialization parameters** and reuse them in `switch_to_training_preset()`.

---

## Key Files Updated

- `trainer.py`
- `preprocess.py`
- `preprocess_lyrics.py`
- `training_handlers.py`
- `training.py`
- `handler.py`
- `__init__.py`

---

## User Impact

- **More robust training startup** on constrained GPUs
- **Fewer device/dtype mismatch crashes** during preprocessing and training
- **Clearer and earlier error reporting** for unsupported training configurations
- **Graceful behavior** when optional logging dependencies are missing

---

Copilot review: response and rebuttals

_Pre-check note_: pre-existing mojibake / encoding artifacts are **out of scope** for this commit review and should be ignored. That said, **symbol-based dynamic string matching is fragile** and should be minimized where possible.

---

## **High Severity**

_Copilot comment_: **trainer.py (lines 230–237)** — LoRA fail-fast with quantization enabled should auto-recover.  
_Response_: **Rejected (not a bug for this commit)**. The UI path already attempts automatic recovery via `switch_to_training_preset()`. Trainer-level fail-fast behavior is **intentional defensive behavior** for non-UI callers.

---

_Copilot comment_: **preprocess.py (lines 127–160)** — `next(model.parameters())` may fail if the model has no parameters.  
_Response_: **Accepted (hardening follow-up)**. Not an observed regression, but we should add a **guarded fallback** for `StopIteration` / uninitialized-model edge cases.

---

_Copilot comment_: **handler.py (lines 534–566)** — `last_init_params` reliance could cause silent failure.  
_Response_: **Rejected**. Failure is **explicit** (`False` return + status message), and `last_init_params` is persisted only after **successful initialization by design**.

---

## **Medium Severity**

_Copilot comment_: **training_handlers.py (lines 549–694)** — Low-VRAM warnings are not actionable enough.  
_Response_: **Accepted (UX improvement, non-blocking)**. We should append **concrete mitigation tips** (e.g. smaller batch size, gradient accumulation, disabling optional components).

---

_Copilot comment_: **preprocess_lyrics.py (lines 13–20)** — Assumes single-device text encoder.  
_Response_: **Pre-existing architectural limitation** (single-device / offload lifecycle). Not a regression introduced by this patch.

---

_Copilot comment_: **handler.py (lines 738–780)** — Missing `model_name` validation may cause silent failures.  
_Response_: **Pre-existing behavior, low risk in this context**. Call sites use fixed internal names; this commit only adjusted **VAE dtype handling on offload**.

---

## **Low Severity**

_Copilot comment_: **training.py (lines 24–427)** — Debug epoch defaults may be inconsistent if toggled mid-session.  
_Response_: **Rejected (not a regression)**. This is **config-at-load behavior**, not a new runtime toggle path introduced by this commit.

---

_Copilot comment_: **trainer.py (lines 285–296)** — TensorBoard warning should suggest install command.  
_Response_: **Accepted (minor polish)**. Add `pip install tensorboard` guidance to the warning text.

---

## **Regression Checks Summary**

_Copilot comment_: Reliance on `last_init_params`, user-visible warnings, and multi-device assumptions need follow-up.  
_Response_: **Partially accepted**.  
- Actionable warning text and preprocess parameter-guard hardening are valid follow-ups.  
- `last_init_params` silent-failure concern is **not valid**.  
- Multi-device concerns are **pre-existing and out of scope** for this patch.

---

### DIFF
--- a/acestep/gradio_ui/events/training_handlers.py
+++ b/acestep/gradio_ui/events/training_handlers.py
@@ -12,6 +12,7 @@ import gradio as gr
 
 from acestep.training.dataset_builder import DatasetBuilder, AudioSample
 from acestep.debug_utils import debug_log_for, debug_start_for, debug_end_for
+from acestep.gpu_config import get_global_gpu_config
 
 
 def create_dataset_builder() -> DatasetBuilder:
@@ -549,6 +550,37 @@ def start_training(
         yield "� Model not initialized. Please initialize the service first.", "", None, training_state
         return
     
+    # Training preset: LoRA training must run on non-quantized DiT.
+    if getattr(dit_handler, "quantization", None) is not None:
+        gpu_config = get_global_gpu_config()
+        if gpu_config.gpu_memory_gb <= 0:
+            yield (
+                "WARNING: CPU-only training detected. Using best-effort training path "
+                "(non-quantized DiT). Performance will be sub-optimal.",
+                "",
+                None,
+                training_state,
+            )
+        elif gpu_config.tier in {"tier1", "tier2", "tier3", "tier4"}:
+            yield (
+                f"WARNING: Low VRAM tier detected ({gpu_config.gpu_memory_gb:.1f} GB, {gpu_config.tier}). "
+                "Using best-effort training path (non-quantized DiT). Performance may be sub-optimal.",
+                "",
+                None,
+                training_state,
+            )
+
+        yield "Switching model to training preset (disable quantization)...", "", None, training_state
+        if hasattr(dit_handler, "switch_to_training_preset"):
+            switch_status, switched = dit_handler.switch_to_training_preset()
+            if not switched:
+                yield f"ï¿½ {switch_status}", "", None, training_state
+                return
+            yield f"ï¿½ {switch_status}", "", None, training_state
+        else:
+            yield "ï¿½ Training requires non-quantized DiT, and auto-switch is unavailable in this build.", "", None, training_state
+            return
+
     # Check for required training dependencies
     try:
         from lightning.fabric import Fabric
@@ -603,9 +635,23 @@ def start_training(
         # Collect loss history
         step_list = []
         loss_list = []
+        training_failed = False
+        failure_message = ""
         
         # Train with progress updates using preprocessed tensors
         for step, loss, status in trainer.train_from_preprocessed(tensor_dir, training_state):
+            status_text = str(status)
+            status_lower = status_text.lower()
+            if (
+                status_text.startswith("âŒ")
+                or status_text.startswith("❌")
+                or "training failed" in status_lower
+                or "error:" in status_lower
+                or "module not found" in status_lower
+            ):
+                training_failed = True
+                failure_message = status_text
+
             # Calculate elapsed time and ETA
             elapsed_seconds = time.time() - start_time
             time_info = f"⏱️ Elapsed: {_format_duration(elapsed_seconds)}"
@@ -648,6 +694,12 @@ def start_training(
         
         total_time = time.time() - start_time
         training_state["is_training"] = False
+        if training_failed:
+            final_msg = f"{failure_message}\nElapsed: {_format_duration(total_time)}"
+            logger.warning(final_msg)
+            log_lines.append(failure_message)
+            yield final_msg, "\n".join(log_lines[-15:]), loss_data, training_state
+            return
         completion_msg = f"� Training completed! Total time: {_format_duration(total_time)}"
         
         logger.info(completion_msg)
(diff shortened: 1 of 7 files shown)


1larity:
@ChuxiJ Mojibake issue should really be addressed, multiple places in existing code are fragile due to use of Mojibake in string compares.

Repo-wide scan found these fragile status-string comparisons:

training_handlers.py (line 646)
status_text.startswith("âŒ") (mojibake literal compare)
training_handlers.py (line 647)
status_text.startswith("❌")
training_handlers.py (line 648)
"training failed" in status_lower
training_handlers.py (line 649)
"error:" in status_lower
training_handlers.py (line 650)
"module not found" in status_lower
api_server.py (line 1464)
audio_codes.startswith("❌")
llm_inference.py (line 413)
status_msg.startswith("❌")
label_utils.py (line 15)
not codes_string.startswith("❌")
results_handlers.py (line 1110)
not alignment_report.startswith("\n⚠️")
results_handlers.py (line 1137)
not alignment_report.startswith("\n⚠️")

Also relevant (not compare, but encoding-damaged text): many mojibake UI strings in training_handlers.py (for example around lines 45, 353-356, 540+, 703, 715).

locale/encoding can impact this:

If text is decoded with wrong codepage (for example cp1252 vs UTF-8), emoji markers become mojibake and compares fail.
Unicode variation selectors (for example ⚠ vs ⚠️) can fail startswith even when visually similar.

Locale itself is less the core issue than I/O encoding + normalization + codepage at boundaries.

acestep/gradio_ui/events/training_handlers.py
```diff
@@ -551,6 +550,37 @@ def start_training(
...
+                "",
+                None,
+                training_state,
+            )
+
+        yield "Switching model to training preset (disable quantization)...", "", None, training_state
+        if hasattr(dit_handler, "switch_to_training_preset"):
+            switch_status, switched = dit_handler.switch_to_training_preset()
+            if not switched:
+                yield f"ï¿½ {switch_status}", "", None, training_state
```
ChuxiJ, line 577: There seems to be an issue with the demo on Gradio, as some people have reported it. I made changes halfway through. Are you sure you want to continue using this character?

ChuxiJ: use❌ or ⚠️ or ✅

ChuxiJ (approved):

PR #358, opened 2026-02-08
fix: avoid calling Module.to() on AffineQuantizedTensor parameters

jayvenn21:
### Description

This PR addresses a regression when moving quantized models to CUDA that results in:

NotImplementedError: AffineQuantizedTensor dispatch:
attempting to run unimplemented operator aten._has_compatible_shallow_copy_type

#### Root cause

During device transfer, `Module.to()` is applied recursively to all parameters.  
When this reaches `AffineQuantizedTensor` parameters (from `torchao`), PyTorch calls
`aten._has_compatible_shallow_copy_type`, which is not implemented for this tensor type.

This causes the model transfer to fail before inference begins.

#### Fix

This change avoids calling `Module.to()` on parameters backed by
`AffineQuantizedTensor`, leaving quantized weights untouched during device transfer.

This matches common handling patterns for quantized models and avoids invoking
unsupported PyTorch internals.

#### Notes

- Issue is reproducible on CUDA with quantized models
- Confirmed by multiple users across different GPU / PyTorch versions
- No `torchao` version currently implements the missing operator, so this avoids the failure path entirely

Let me know if you’d prefer this logic to live closer to the device-transfer utility or behind a more explicit quantization check.

## Summary by CodeRabbit

* **New Features**
  - Added an option to skip moving quantized parts while still transferring non-quantized components, and extended traversal to discover module-like attributes for consistent device transfers.

* **Bug Fixes**
  - Prevented errors when performing device operations on models containing quantized parameters/buffers by avoiding unsafe moves of quantized components.


ChuxiJ (approved):

ChuxiJ:
please solve the conflicts and use `uv sync --reinstall` to test

PR #401, opened 2026-02-09
Enhance MPS LoRA training by enabling gradient checkpointing in DiT training path

riversedge:
### **_This should get peer review before merge because it touches core training control flow and wrapper interactions._**

It looks to be running correctly on my M4 Max 64 GB and MPS improvements seems promising, but area of change just makes me a little cautious.   The reduction of memory pressure and the GPU utilization improvement and throughput increases though are great if this didn't break anything.

**Primary change:**
Add explicit grad-carrying fallback input for checkpointed forward (xt.requires_grad_(True)), since this DiT model does not implement the LM-style embedding hook path used by PEFT (get_input_embeddings).

**Supporting changes required for stability:**
PEFT/DiT compatibility guard around enable_input_require_grads.
Decoder unwrap before reinjecting LoRA after failed/partial wrapped states.
MPS AMP + dtype stabilization (ensure trainable/optimizer params remain fp32 where required).
Non-finite gradient guard to skip unstable steps instead of crashing.
Logging cleanup for expected DiT-specific checkpointing behavior.
Notes:

Intended behavior change is primarily for MPS training robustness; non-MPS paths should be functionally unchanged except for safer guards.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of model wrappers and gradient operations during training.
  * Added safeguards against non-finite gradients.

* **New Features**
  * Enhanced gradient checkpointing and memory optimization configuration.
  * Implemented FP32 precision enforcement for model parameters.
  * Added automatic detection and reporting of training memory features, including gradient checkpointing and cache usage.


riversedge:
FWIW I successfully completed a 30 min source audio LoRA training run with this code and tested out several songs prompts.   While certain parts of the output quality from that run were perhaps lower than desired, it was clear that the training had worked and adopted the source song styles successfully.   Each epoch run took apps 34s on MPS Mac M4 Max, memory footprint was stable, training did converge nicely (see below).   Total run about 8 hrs for 1000 epoch (vs mainline branch taking about 80 hours).

Screenshot 2026-02-09 at 10 16 19 PM

ChuxiJ (approved):

scf4:
How comparable are LoRAs trained on mps vs same exact run (same seed/data/params/dataset.json) on cuda? (Assuming training is ~deterministic)

(Basically could this be used for quick/cheap local iteration before "real" runs)

PR #404, opened 2026-02-09
(fix)Add CORS support for direct open browser-based frontends (studio.html)

tonyjohnvan:
Problem
-------
Opening ui/studio.html via file:// protocol and calling the API at http://localhost:8001 fails with:

  "Access to fetch at '…/release_task' from origin 'null' has been
   blocked by CORS policy: Response to preflight request doesn't pass
   access control check"

The browser sends a preflight OPTIONS request for the cross-origin POST, but neither the standalone FastAPI server (api_server.py) nor the Gradio-mounted API routes (api_routes.py) return the required Access-Control-Allow-Origin header.

Changes
-------
1. acestep/api_server.py
   - Import CORSMiddleware from fastapi.middleware.cors.
   - Add CORSMiddleware to the standalone FastAPI app immediately after creation (before the app starts, so add_middleware() works normally).

2. acestep/gradio_ui/api_routes.py
   - Import CORSMiddleware from fastapi.middleware.cors.
   - Add _CORS_KWARGS dict with the shared CORS configuration.
   - Add _add_cors_middleware() for the normal pre-launch path (used by setup_api_routes_to_app).
   - Add _add_cors_middleware_post_launch() that patches the compiled middleware_stack directly, because Gradio's demo.launch() starts the Starlette app before setup_api_routes() is called, and Starlette raises "Cannot add middleware after an application has started" if add_middleware() is used post-launch.
   - Wire both setup functions to their respective CORS helpers.

CORS policy (applied uniformly in both files)
----------------------------------------------
  allow_origins       : ["null", "http://localhost", "http://127.0.0.1"]
  allow_origin_regex  : ^https?://(localhost|127\.0\.0\.1)(:\d+)?$
  allow_methods       : ["GET", "POST", "OPTIONS"]
  allow_headers       : ["Content-Type", "Authorization"]
  allow_credentials   : not set (defaults to False)

Benefits
--------
- studio.html (opened via file://, origin "null") can now call the API.
- Any local dev server on localhost or 127.0.0.1 (any port) also works.
- The Gradio + API path no longer crashes on startup.

Risks & mitigations
--------------------
- Origin "null" is allowed. The "null" origin is sent by file:// pages, sandboxed iframes, and some redirect flows. A malicious local HTML file could call the API if opened in the same browser. Mitigation: the API is bound to localhost by default, and API key auth (--api-key) is available for sensitive deployments.
- If the server is exposed to the network (0.0.0.0 or --share), only localhost origins are allowed — external sites cannot call the API. However, the "null" origin exception still applies to any file:// page on the user's machine.
- The post-launch middleware patching (_add_cors_middleware_post_launch) accesses app.middleware_stack directly, which is a Starlette internal. This could break if Starlette refactors its middleware compilation. Mitigation: guarded by an `if app.middleware_stack is not None` check and falls back to the normal add_middleware path otherwise.
- allow_credentials is intentionally omitted (False). studio.html uses plain JSON fetch without cookies. If cookie-based auth is added later, this will need revisiting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved cross-origin resource sharing (CORS) issues, enabling browser-based frontends to access the API from localhost and file:// protocol origins.

### DIFF
--- a/acestep/api_server.py
+++ b/acestep/api_server.py
@@ -41,6 +41,7 @@ except ImportError:  # Optional dependency
     load_dotenv = None  # type: ignore
 
 from fastapi import FastAPI, HTTPException, Request, Depends, Header
+from fastapi.middleware.cors import CORSMiddleware
 from pydantic import BaseModel, Field
 from starlette.datastructures import UploadFile as StarletteUploadFile
 
@@ -2143,6 +2144,16 @@ def create_app() -> FastAPI:
             executor.shutdown(wait=False, cancel_futures=True)
 
     app = FastAPI(title="ACE-Step API", version="1.0", lifespan=lifespan)
+
+    # Enable CORS for browser-based frontends (e.g. studio.html opened via file://)
+    # Restricted to localhost origins and the "null" origin (file:// protocol)
+    app.add_middleware(
+        CORSMiddleware,
+        allow_origins=["null", "http://localhost", "http://127.0.0.1"],
+        allow_origin_regex=r"^https?://(localhost|127\.0\.0\.1)(:\d+)?$",
+        allow_methods=["GET", "POST", "OPTIONS"],
+        allow_headers=["Content-Type", "Authorization"],
+    )
 
     # Mount OpenRouter-compatible endpoints (/v1/chat/completions, /v1/models)
     from acestep.openrouter_adapter import create_openrouter_router
--- a/acestep/gradio_ui/api_routes.py
+++ b/acestep/gradio_ui/api_routes.py
@@ -10,6 +10,7 @@ from typing import Any, Dict, List, Optional
 from uuid import uuid4
 
 from fastapi import APIRouter, HTTPException, Request, Depends, Header
+from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import FileResponse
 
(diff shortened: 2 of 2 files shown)


tonyjohnvan:
@ChuxiJ I took a look into the ui and the api folder and found the original simple html won't be working by  "click to open" mechanism. the API backend was not setup CORS for good, this temporally fix is allowing direct html file open to access the API features, more safer approach is to let user `cd ./ui && python3 -m http.server <PORT>` to host the HTML file in a simple web server. I did this change anyway since the doc allow user to directly open the file for simpler UI and API access without mentioning about the web server. But it's really up to you to pull the trigger to determine should we go with CORS or web server
 
(your IDE might have build in web server for you to "click to open" the HTML file directly? I used to use Intelij IDEA I know they does like that)

ChuxiJ:
We don’t need to develop or design this part for now.
Let’s focus our recent development on device compatibility and the stable release when submitting PRs.
We’ll handle UI optimizations later in a targeted way.

By then, we’ll support more features to better cover workflows like Cover, Remix, and Edit, along with project refactoring and more complete tutorials.
Let’s set aside these issues that aren’t part of our current focus for the time being.

PR #493, opened 2026-02-12
fix: phase-aware max_new_tokens to fix misleading progress bar stopping early

agorevski:
The progress bar showed generation stopping at some 66%% (975/1475) because max_new_tokens always added a +500 buffer regardless of generation phase.

In the two-phase architecture, the codes phase only generates audio codes (CoT is already in the prompt), and the constrained decoder forces EOS at exactly target_codes. The +500 buffer was unnecessary and inflated the progress bar denominator.

Changes:
- Add _compute_max_new_tokens() helper with phase-aware calculation:
  - codes phase: target_codes + 10 (small buffer, constrained decoder forces EOS)
  - cot phase: target_codes + 500 (larger buffer for metadata overhead)
- Replace all 5 inline calculations across vllm, pt, and mlx backends
- Add 10 unit tests covering all calculation paths and regression case

## Summary by CodeRabbit

* **Refactor**
  * Centralized token-budget calculation across inference backends with phase-aware buffers (smaller for code, larger for chain-of-thought), duration clamping, and consistent capping to prevent oversized generations.

* **Tests**
  * Added comprehensive unit tests covering code/CoT phases, duration bounds, fallback behavior, capping, edge cases, and a regression check for progress reporting.

### DIFF
--- a/acestep/api_server.py
+++ b/acestep/api_server.py
@@ -1585,11 +1585,30 @@ def create_app() -> FastAPI:
 
                 # Build GenerationConfig - default to 2 audios like gradio_ui
                 batch_size = req.batch_size if req.batch_size is not None else 2
+
+                # Resolve seed(s) from req.seed into List[int] for GenerationConfig.seeds
+                resolved_seeds = None
+                if not req.use_random_seed and req.seed is not None:
+                    if isinstance(req.seed, int):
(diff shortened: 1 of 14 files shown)


ChuxiJ:
please solve conflicts

PR #507, opened 2026-02-13
Potential fix for code scanning alert no. 35: Uncontrolled data used in path expression

schneidergithub:
Potential fix for [https://github.com/ace-step/ACE-Step-1.5/security/code-scanning/35](https://github.com/ace-step/ACE-Step-1.5/security/code-scanning/35)

At a high level, we want to constrain and validate any filesystem paths derived from user input before using them. In this case, the risky usage happens in `export_lora` in `acestep/gradio_ui/events/training_handlers.py`, where both the source (`lora_output_dir`) and destination (`export_path`) paths are derived from user input. The fix is to (1) normalize those paths, (2) enforce that they stay within a defined safe root directory (for example, the current working directory or a project root), and (3) reject any absolute or traversal paths that attempt to escape that root.

The single best fix with minimal behavior change is to add a small helper function (inside `training_handlers.py`) that takes a user-supplied path and a configured root directory, normalizes it with `os.path.abspath`, and ensures the normalized path is inside that root using `os.path.commonpath`. We then apply this helper to both `lora_output_dir` and `export_path` inside `export_lora` before using them. If validation fails, we return a clear error message instead of proceeding. We should keep the existing logic (preferring `final` over `checkpoints`, choosing the latest epoch, copying directories) untouched.

Concretely:

- In `acestep/gradio_ui/events/training_handlers.py`, just above `export_lora`, define a helper like `_safe_join(base_root: str, user_path: str) -> Optional[str]` that:
  - Strips the input, rejects empty strings.
  - Joins `base_root` and the user path.
  - Normalizes to an absolute path via `os.path.abspath`.
  - Uses `os.path.commonpath([base_root, abs_path]) == base_root` to assert the path is inside `base_root`.
  - Returns the safe absolute path or `None` on failure.
- Decide on a root. With the limited context, the least intrusive safe default is the current working directory, e.g. `SAFE_ROOT = os.path.abspath(os.getcwd())`, defined once at module level. That constrains all training I/O to the process's working tree.
- In `export_lora`:
  - Compute `safe_root = SAFE_ROOT`.
  - Validate `lora_output_dir` with `_safe_join(safe_root, lora_output_dir)`; if invalid, return `"❌ Invalid LoRA output directory"`.
  - Use that validated `safe_lora_output_dir` when constructing `final_dir` and `checkpoint_dir`.
  - Similarly, validate `export_path` using `_safe_join(safe_root, export_path)`. If invalid, return a clear error. Use the validated path for `os.makedirs`, `shutil.rmtree`, and `shutil.copytree`.
- This addresses all 59 variants because the CodeQL taint flow from many Gradio widgets eventually ends up at the same sink (`final_dir` in `export_lora`); by sanitizing here, all those flows are cut off.

No changes are needed in `__init__.py` or `interfaces/training.py` for this specific sink; we only tighten how paths are handled right before filesystem operations.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

## Release Notes

* **Bug Fixes**
  * Improved validation of file paths during training operations
  * Added safeguards to prevent invalid export paths and enhance filesystem security
  * Enhanced stability of training-related file operations

### DIFF
--- a/acestep/gradio_ui/events/training_handlers.py
+++ b/acestep/gradio_ui/events/training_handlers.py
@@ -17,6 +17,11 @@ from acestep.training.dataset_builder import DatasetBuilder, AudioSample
 from acestep.debug_utils import debug_log_for, debug_start_for, debug_end_for
 from acestep.gpu_config import get_global_gpu_config
 
+# Define a safe root directory for all training-related filesystem operations.
+# This limits user-supplied paths (for checkpoints, exports, etc.) to stay
+# within the server's working tree, preventing directory traversal outside it.
+SAFE_TRAINING_ROOT = os.path.abspath(os.getcwd())
+
 
 def create_dataset_builder() -> DatasetBuilder:
     """Create a new DatasetBuilder instance."""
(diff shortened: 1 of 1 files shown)


schneidergithub:
@ChuxiJ This fix will probably work, but could cause issues with existing setups relying on hard-coded values.  Regardless, the codeQL alerts are showing some critical / high alerts that can be fixed pretty quickly with AI refactors.  I just wanted to give you a quick update before I take off.  Good night!

PR #579, opened 2026-02-15
Training UI: add “external API” autolabel option + provider override

robustini:
Summary
This PR introduces a lightweight, opt-in “external provider” hook for single-sample autolabeling in the training UI. The goal is to keep the core labeling workflow unchanged by default, while allowing users to route labeling to a user-configured external API server when they explicitly choose to do so.

Motivation
Some users and teams already run their own annotation services (self-hosted or enterprise) and want a clean integration point without forking the project or adding heavyweight dependencies to the default path. This change provides that integration point while keeping the repository provider-agnostic.

Behavior and scope
Default behavior remains unchanged when no external provider is selected or configured. The external path is explicitly opt-in and intended as an integration hook rather than an endorsement of any specific model or service. Output quality depends entirely on the chosen external backend.

Configuration
External provider selection can be controlled via the UI action and/or an environment variable (e.g. ACESTEP_AUTOLABEL_PROVIDER). The external endpoint can be configured using provider-specific settings as documented in code comments.

Test plan
I verified that the training UI still works identically without any provider configuration, and that the external autolabel action populates the relevant fields when the external provider is enabled. I also verified that failures such as an unreachable endpoint are handled cleanly without corrupting dataset state.

Alternatives
I intentionally avoided hard-wiring a paid vendor API into the default experience, and I avoided shipping a specific open model as a required dependency, to keep upstream maintenance and the default runtime lightweight. Keeping everything fully out-of-tree was considered, but a small upstream hook reduces duplicated forks and keeps integrations consistent.

Future work
If desired, a follow-up can add a small provider registry, structured output validation, and/or optional batch labeling, all behind opt-in gating.

Naming
If preferred by maintainers, I can rename the UI button to a more generic label such as “Autolabel with external API server (experimental)” and adjust any wording to match project conventions.

## Release Notes

* **New Features**
  * Added Music-Flamingo autolabel functionality to the training tab, providing an alternative AI-powered method to label audio samples with a dedicated button.
  * Standardized status and error messages across the training interface with visual indicators (✅, ❌, ⏱️, ⚠️, ℹ️) for improved clarity and consistency.

### DIFF
--- a/acestep/training/music_flamingo_autolabel.py
+++ b/acestep/training/music_flamingo_autolabel.py
@@ -56,8 +56,11 @@ DEFAULT_PROMPT_DESCRIBE_FULL = (
     "Write a single-paragraph track card (no line breaks), 90-120 words. Include: genre + overall vibe + broad instrumentation/production + vocals (if present). Add metadata inline exactly like: (Key: X; BPM: Y; Meter: Z) using unknown if unsure. Do NOT quote lyrics or include example lines. For themes, use one short paraphrase sentence only. Avoid: verse/chorus/drop structure, stereo imaging, EQ/compression/mastering jargon, and overly specific vocal type labels (no \"tenor/baritone\"). Output only the paragraph."
 )
 
-DEFAULT_PROMPT_LYRICS = "Extract the lyrics"
-
+DEFAULT_PROMPT_LYRICS = (
+    "Return ONLY the sung/spoken words as plain text. "
+    "No analysis, no commentary, no song structure labels (verse/chorus/bridge), and no timestamps. "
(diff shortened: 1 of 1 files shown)


robustini:
I’m going to close this PR for now. Although the integration works in principle, the results are too inconsistent for an autolabeling pipeline and often need manual cleanup. On top of that, latency is highly variable (with frequent ~120s+ responses/timeouts depending on queue/cold-start), which makes batch labeling brittle, and the dependency on a public hosted endpoint is difficult to rely on for repeatable workflows.
I’m happy to revisit this approach with a more deterministic local ASR backend (e.g., faster-whisper/whisper.cpp) or a self-hosted alternative.
Best,
Marco

PR #582, opened 2026-02-15
Add lycoris lokr load and load lora api

sdbds:
Add lycoris lokr load and load lora 、training api

## Summary by CodeRabbit

* **New Features**
  * LoRA management endpoints: load, unload, toggle, scale, and status.
  * Dataset operations: scan, load, auto-label (sync/async), preprocess (sync/async), save, query, and update samples.
  * Training orchestration: start standard and LoKR training, monitor progress, stop training, export trained weights, and TensorBoard integration.
  * Background task status for auto-labeling and preprocessing; improved training config options (FP8, gradient checkpointing).

* **Minor Improvements**
  * Safer persistence and logging; improved runtime resource handling and restore.

### DIFF
--- a/acestep/gradio_ui/events/__init__.py
+++ b/acestep/gradio_ui/events/__init__.py
@@ -135,6 +135,37 @@ def setup_event_handlers(demo, dit_handler, llm_handler, dataset_handler, datase
         outputs=[generation_section["lora_status"]]
     )
     
+    # ========== Auto Checkbox Handlers ==========
+    _auto_field_map = {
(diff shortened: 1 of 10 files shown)


ChuxiJ:
I think we need to separate the training-related APIs into a distinct module.
For example, it would be appropriate to create a separate file like train_api_service.py for them,
so as to avoid making the original file too large.

ChuxiJ:
Furthermore, the CodeQL scanning process failed to complete successfully.

PR #671, opened 2026-02-22
fix(generation): add timeout, progress fallback, and VRAM pre-flight …

### LINKED ISSUE #146: Inference stuck at 50% (Phase 2) | RTX 5060 Ti 16GB | Request for Stop/Abort Button
Image

Describe the bug 

During a long-duration generation (235 seconds) with a Batch Size of 2, the process hangs indefinitely at 50.0% (Phase 2: Generating audio codes). The terminal shows a successful prefill but the decoding stage never progresses beyond 0/2.

System Specs:

GPU: NVIDIA GeForce RTX 5060 Ti (16GB VRAM)

Environment: Windows 10 Portable Package (python_embeded)

Models used: DiT acestep-v15-turbo + LM acestep-5Hz-lm-1.7B

Steps to Reproduce:

Set Audio Duration to 235s.

Set Batch Size to 2.

Click Generate Music.

Observe the UI hang at 50% and terminal showing Decode=83tok/s with no progress.

Missing Feature: There is currently no Stop/Cancel button in the Gradio UI. The only way to interrupt a hung process is to force-close the terminal, which can lead to corrupted checkpoints or temporary files.

Requested Improvements:

Implementation of an Interrupt/Stop button for inference.

Investigation into why long-duration batches (Batch Size > 1) cause decoding to hang on 16GB VRAM cards despite being within the "Tier 5" recommendation.

Image

Image

Uni404x64:
…— fixes #146

Addresses three root causes of the 50% UI freeze on mid-tier GPUs:

1. **Synchronous deadlock** (`generate_music_execute.py`)
   - Wrapped `service_generate()` in a daemon thread with a configurable wall-clock timeout (default 600 s, override via `ACESTEP_GENERATION_TIMEOUT`).
   - Safe env-var parsing: non-numeric values and <= 0 fall back to 600 s with a warning instead of crashing at import time or timing out every generation.
   - Calls `torch.cuda.empty_cache()` before raising `TimeoutError` to recover fragmented VRAM for subsequent attempts.
   - Logs orphaned `service-generate` thread count on timeout for operator visibility.
   - Uses `except BaseException` in the worker to ferry `SystemExit` and friends across the thread boundary without swallowing them.
   - Defensive guard on `_result["outputs"]` prevents a raw `KeyError` if the thread exits without populating either dict.

2. **Estimator cold-start** (`progress.py`)
   - `_start_diffusion_progress_estimator` no longer returns `(None, None)` when no timing history exists; falls back to `2.5 s/step x batch_size` so the progress bar always animates on first run.
   - Self-calibrates after the first successful generation via existing mechanism.

3. **VRAM pre-flight check** (`generate_music.py`)
   - `_vram_preflight_check()` runs before `service_generate`, using the project-standard `get_effective_free_vram_gb()` + `DIT_INFERENCE_VRAM_PER_BATCH` constants from `gpu_config.py`.
   - Detects CFG mode via `guidance_scale > 1.0` (two DiT passes = 2x activations).
   - Scales estimate with duration (5 Hz latent rate).
   - Returns a clean, actionable error payload instead of a silent hang.
   - No-ops on non-CUDA backends; safe if `get_effective_free_vram_gb()` raises.

All logger calls use loguru (`from loguru import logger`) consistent with the rest of the handler package.

Tested: Windows 11 / Ryzen 9 5900X / RTX 3070 Ti (8 GB). Batch 2, 3:01 tracks, 100.78 s total. DiT phase 12.8 s. No hang, progress bar animated throughout.

## Summary by CodeRabbit

* **Improvements**
  * Added GPU memory preflight checks to detect and report insufficient VRAM before generation starts.
  * Implemented generation timeout protection that stops stalled jobs, surfaces errors, and ensures progress cleanup.
  * Enhanced progress estimation with a conservative fallback for cold starts and clearer logging for progress and headroom.


ChuxiJ:
please solve comments

PR #815, opened 2026-03-11
feat: add chunk_mask_mode parameter for explicit vs auto mask control

ChuxiJ:
Introduce a chunk_mask_mode option ("explicit" | "auto") that flows from the HTTP API through GenerationParams down to the conditioning pipeline.

- "explicit": uses the 0/1 chunk mask derived from the repainting range
  (existing behavior, Mask Control: true in prompt).
- "auto": fills the chunk mask tensor with 2.0 so the model decides
  mask boundaries itself (Mask Control: false in prompt).

Default is "auto" which lets the model handle segmentation autonomously.

## Summary by CodeRabbit

* **New Features**
  * Added configurable chunk masking mode to music generation. Users can now specify masking behavior with 'auto' (default) for automatic masking or 'explicit' mode for explicit control during generation. This enhancement provides greater flexibility and precision in customizing music generation workflows according to specific masking requirements, use cases, and preferences.

### DIFF
--- a/acestep/api/http/release_task_models.py
+++ b/acestep/api/http/release_task_models.py
@@ -62,6 +62,7 @@ class GenerateMusicRequest(BaseModel):
         description="User-provided audio semantic codes string for code-control generation. When non-empty, skips LM code generation.",
     )
     task_type: str = "text2music"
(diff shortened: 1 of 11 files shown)


garibovallahverdi:
fg

PR #816, opened 2026-03-11
feat: add step-level source injection and boundary blending for repaint

ChuxiJ:
For repaint tasks, the diffusion loop now enforces non-repaint regions at every step by replacing them with the appropriately-noised source latent (matching ACE-Step 1.0's approach). After the loop, a configurable crossfade blend smooths the transitions at repaint boundaries.

This eliminates error accumulation in preserved regions and removes audible seams at repaint edges. Non-repaint tasks (text2music, cover) are completely unaffected.

Made-with: Cursor

## Summary by CodeRabbit

* **New Features**
  * Targeted audio repainting with per-frame repaint masks for selective region edits
  * Smooth latent boundary blending with configurable crossfade frames and injection ratio
  * Waveform-level splicing after decoding with optional crossfade duration
  * New user controls: Repaint Mode (conservative/balanced/aggressive) and Repaint Strength slider
  * New request/options: latent crossfade frames, wav crossfade seconds, mode, and strength

* **Tests**
  * Added comprehensive unit tests for repaint masking, step injection, boundary blending, waveform splicing, and config mapping


Kwisss:
~I don't know if it's inside the scope of this pr, but how is "audio_cover_strength" related in this process?
both on main and with this current pr, the audio inside the repaint region nulls with a setting higher then "0.0".
Not to take away that this is great pr, it makes the repaint functionality a lot better.~

Found my answers, it's certainly outside the scope of this pr.

PR #828, opened 2026-03-13
Detect missing model weights before skipping downloads

1larity:
## Summary
- guard against treating partially present checkpoints as fully available
- verify required model weights exist before skipping downloader work
- add focused downloader regression tests for missing-weights cases

## Validation
- uv run python -m unittest acestep.model_downloader_test

## Release Notes

* **Bug Fixes**
  * Improved model detection to validate actual weight files rather than directory existence alone.
  * Added support for multiple model weight file formats, including Diffusers-style variants.
  * Enhanced validation to ensure model component completeness.

* **Tests**
  * Added comprehensive tests for model existence and component validation logic.

### DIFF
--- a/.claude/skills/acestep-simplemv/SKILL.md
+++ b/.claude/skills/acestep-simplemv/SKILL.md
@@ -123,9 +123,10 @@ When sending MV to chat platforms, use `--max-size` to auto-compress:
 
 | Platform | Limit | Recommended `--max-size` |
 |----------|-------|--------------------------|
-| WhatsApp | 16MB | 15 |
-| Discord (free) | 25MB | 24 |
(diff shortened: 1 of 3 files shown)


1larity:
A better approach is capability-based validation, with small helpers like “usable DiT checkpoint”, “usable VAE checkpoint”, “usable text encoder checkpoint”, and “usable LM checkpoint”, then compose those for ensure_main_model() / service init.

So instead of  answering “does this directory contain a set of a few filenames (and variations on filenames)?” we answer  “can the app successfully use this model set?”

I can put in a check for the current VAE file, but maybe it is better to raise an issue to expand this to be capability based.

Issue #460, opened 2026-02-11
Windows model init fails with os error 1455 (paging file too small) and retries attention backends, amplifying memory pressure


1larity:
## Severity
High reliability issue (service initialization can fail completely on Windows)
Appears to be caused by repeated model inits by user.
## Summary
Model initialization fails on Windows with:

`OSError: The paging file is too small for this operation to complete. (os error 1455)`

The failure occurs inside `safe_open(...model.safetensors...)` during `AutoModel.from_pretrained(...)`, and init then continues retrying additional attention implementations (`flash_attention_2`, `sdpa`, `eager`), which appears to increase memory pressure further.

## Environment
- OS: Windows
- RAM: 32 GB
- Page file: 60 GB
- Model: `acestep-v15-turbo`
- Path: `checkpoints/acestep-v15-turbo/model.safetensors`

## Repro
1. Start UI.
2. Initialize service with `acestep-v15-turbo` (auto device).
3. Observe failure during model load.

## Actual
Initialization fails with:
`RuntimeError: Failed to load model with attention implementations ['flash_attention_2', 'sdpa', 'eager']: ... (os error 1455)`

Stack excerpt points to:
- `acestep/handler.py` in `initialize_service`
- `transformers` `AutoModel.from_pretrained`
- `safe_open(checkpoint_files[0], framework="pt")`

## Expected
- Either successful model load, or fast/clear failure with actionable guidance.
- On memory-commit failure (`1455`), init should not continue additional backend retries that can worsen pressure.

## Suggested fix
1. Detect Windows memory-commit failures (`OSError` with `1455`) and fail-fast (skip further attention backend retries).
2. Between failed candidates, aggressively cleanup:
   - `del self.model` (if allocated)
   - `gc.collect()`
   - device cache clear/sync (`_empty_cache`, `_synchronize`)
3. Improve user-facing error message:
   - mention commit-memory exhaustion
   - suggest disabling LM init, disabling flash-attn, restarting process, reducing concurrent memory usage.

jayvenn21:
Looks like one of the related PR's is mine. I'll tackle this issue

1larity:
Cheers, it literally crashed my Linux box, and I had to reboot my Windows box as it became generally unstable :)

Issue #135, opened 2026-02-04
Segmentation fault when initializing 5Hz LM with vLLM backend on Linux with torch 2.10.0


Garry-TI:
Description:

  Environment

  - OS: Ubuntu Linux (kernel 6.8.0-90-generic)
  - Python: 3.11
  - PyTorch: 2.10.0+cu128
  - torchao: 0.16.0 (built from git main)
  - diffusers: 0.36.0
  - transformers: 4.57.6
  - flash-attn: 2.8.3+cu128torch2.10
  - triton: 3.6.0
  - nano-vllm: 0.2.0
  - GPU: 24GB VRAM (Configuration Tier: tier6)
  - CUDA: 12.8

  Issue

  When attempting to initialize the 5Hz LM using the vLLM backend (default), the application crashes with a segmentation fault (exit code 139) immediately after printing the deprecation warning
  'torch_dtype' is deprecated! Use 'dtype' instead!

  Steps to Reproduce

  1. Install ACE-Step-1.5 with torch 2.10.0 on Linux
  2. Run uv run acestep
  3. Click the "Initialize" button in the Gradio UI
  4. Application crashes to shell prompt with no Python traceback

  Expected Behavior

  The 5Hz LM should initialize successfully using the vLLM backend.

  Actual Behavior

  The process crashes with exit code 139 (segmentation fault) during LLM initialization. The crash occurs at the C/C++ level, so no Python traceback is produced.

  Terminal output:
  2026-02-04 16:47:13.979 | INFO     | acestep.llm_inference:initialize:384 - Constrained processor initialized in 2.43 seconds
  [debug]dist_port: 2333
  'torch_dtype' is deprecated! Use 'dtype' instead!
  (Process terminates)

  Minimal reproduction:
  from nanovllm import LLM
  from transformers import AutoTokenizer

  model_path = '/path/to/checkpoints/acestep-5Hz-lm-1.7B'
  tokenizer = AutoTokenizer.from_pretrained(model_path, use_fast=True, trust_remote_code=True)

  llm = LLM(
      model=model_path,
      enforce_eager=False,
      tensor_parallel_size=1,
      max_model_len=4096,
      gpu_memory_utilization=0.34,
      tokenizer=tokenizer,
  )
  # Segmentation fault occurs here

  Exit code: 139 (segfault)

  Workaround

  Using the PyTorch backend instead of vLLM works successfully:
  uv run acestep --backend pt

[...]

PR #830, opened 2026-03-13
Improve Windows LM backend fallback messaging

### LINKED ISSUE #616: nano-vLLM initialization fails, falls back to PT backend and crashes during generation (on Google Colab L4)
**Summary**
I believe I’ve found a regression introduced after commit 6db4465ca0a57c3a57416c4acb73fd5e5dad8986.

On NVIDIA L4 (22GB) with Torch 2.10.0+cu128, newer commits fail to initialize the 5Hz LM using nano-vLLM due to CUDA graph capture errors. ACE-Step then falls back to the PyTorch backend, which later crashes during generation with:

`RuntimeError: Offset increment outside graph capture encountered unexpectedly.`

Pinning to commit 6db4465 (from ~2 days ago) restores correct behavior — vLLM initializes successfully and generation works.

**Environment**
- GPU: NVIDIA L4 (22GB VRAM)
- Runtime: Google Colab
- CUDA detected by ACE-Step: tier6b
- Torch: 2.10.0+cu128
- nano-vLLM installed via project setup
- LM model: acestep-5Hz-lm-1.7B
- DiT config: acestep-v15-turbo

**More details**
During startup of Gradio:
```
Initializing 5Hz LM with model: ..., 
[nanovllm] KV cache allocated ...
❌ Error initializing 5Hz LM:
CUDA error: operation failed due to a previous error during capture
...
torch.AcceleratorError: CUDA error: operation not permitted when stream is capturing
```
which then yields
```
WARNING Falling back to PyTorch backend
5Hz LM initialized successfully using PyTorch backend on cuda
```

and if i go to generate a song:
```RuntimeError: Offset increment outside graph capture encountered unexpectedly.```

**What I tried**
- Restarted Gradio / service multiple times
- Disabled torch.compile/dynamo via env vars at launch:
- TORCHDYNAMO_DISABLE=1
- TORCH_COMPILE_DISABLE=1
- TORCHINDUCTOR_DISABLE_CUDAGRAPHS=1
- Selected vLLM backend in UI
- Tried to force eager mode (enforce_eager=True) to disable CUDA graphs, but it did not resolve the issue — logs still show enforce_eager: False and nano-vLLM still attempts CUDA graph capture / fails.

**To reproduce:**
1. Launch Gradio on Colab with NVIDIA L4
2. Initialize service (DiT + LM enabled)
3. Observe vLLM init failure → PT fallback
4. Generate a song → generation crash

1larity:
## Summary
- introduce shared LM backend compatibility helpers for Windows fallback handling
- improve user-facing messaging when vLLM/Triton is unavailable on Windows
- cover the fallback behavior with focused compatibility tests

fixes #616, #135 partially addresses #460.

## Validation
- uv run python -m unittest acestep.llm_backend_compat_test

## Summary by CodeRabbit

* **New Features**
  * Adds runtime preflight checks for vLLM on Windows+CUDA and auto-falls back to the PyTorch backend with a suppressible warning.

* **User-facing changes**
  * Logs the preflight warning when fallback occurs and appends a note to initialization/status messages; clearer fallback guidance and standardized status tokens for diagnostics.

* **Tests**
  * Added tests covering preflight detection, Windows/CUDA fallback paths, and status message outcomes.

### DIFF
--- a/.claude/skills/acestep-simplemv/SKILL.md
+++ b/.claude/skills/acestep-simplemv/SKILL.md
@@ -123,9 +123,10 @@ When sending MV to chat platforms, use `--max-size` to auto-compress:
 
 | Platform | Limit | Recommended `--max-size` |
 |----------|-------|--------------------------|
-| WhatsApp | 16MB | 15 |
-| Discord (free) | 25MB | 24 |
-| Telegram | 50MB | 48 |
+| WhatsApp | 16MB | 10 |
+| Discord (free) | 25MB | 10 |
+| QQ | 10MB | 10 |
+| Telegram | 50MB | 20 |
 | Slack (free) | 1GB | - |
 
 The compression uses ffmpeg two-pass encoding to achieve the best quality within the size constraint.
--- a/README.md
+++ b/README.md
@@ -107,6 +107,8 @@ Open http://localhost:7860 (Gradio) or http://localhost:8001 (API).
 
 > 📦 **Windows users:** A [portable package](https://files.acemusic.ai/acemusic/win/ACE-Step-1.5.7z) with pre-installed dependencies is available. See [Installation Guide](./docs/en/INSTALL.md#-windows-portable-package).
 
+> 📦 **MacOS users:** A [portable package](https://files.acemusic.ai/acemusic/mac/ACE-Step-1.5.zip) with pre-installed dependencies is available. See [Installation Guide](./docs/en/INSTALL.md#-macos-portable-package).
+
 > 📖 **Full installation guide** (AMD/ROCm, Intel GPU, CPU, environment variables, command-line options): [English](./docs/en/INSTALL.md) | [中文](./docs/zh/INSTALL.md) | [日本語](./docs/ja/INSTALL.md)
 
 ### 💡 Which Model Should I Choose?
--- a/docs/en/INSTALL.md
+++ b/docs/en/INSTALL.md
@@ -10,6 +10,7 @@
 - [Quick Start (All Platforms)](#quick-start-all-platforms)
 - [Launch Scripts](#-launch-scripts)
 - [Windows Portable Package](#-windows-portable-package)
+- [macOS Portable Package](#-macos-portable-package)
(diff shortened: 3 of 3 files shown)


1larity:
The icons are a complete nightmare on Windows, I *think* i got that right, fingers crossed.

1larity:
@coderabbit I addressed the main actionable point in `53f8c35`: when vLLM fails with the Triton-specific error, the final PyTorch fallback status now preserves that explanation instead of overwriting it with only a generic fallback success message. I also updated the regression test to assert that the returned status includes the Triton explanation.

On the cleanup/reset suggestion: I checked that codepath carefully. `_initialize_5hz_lm_vllm()` already performs the available pre-init cleanup in this module before constructing vLLM, including `torch.cuda.empty_cache()` and `_cleanup_torch_distributed_state()`. In the Triton case, the failure occurs during vLLM initialization and we do not currently have a more specific vLLM teardown handle/state in this path to clean up. I don’t see a clearly missing cleanup step here without a concrete stale-state repro, so I’m not adding speculative teardown logic in this PR.

If we do observe a reproducible leaked-state problem after a Triton-init failure, I’m happy to follow up with a targeted cleanup patch.

ChuxiJ (approved):
Re-reviewed after the fix commits. The primary blocker (emoji→ASCII scope creep) has been fully removed — all status messages now use the existing ❌/✅ convention. The PR is now tightly scoped to its stated purpose.

What's good:
- Clean 36-LOC preflight module (`llm_backend_compat.py`) with proper docstrings
- Two-layer defense: preflight check prevents the crash, runtime handler catches it if it still happens
- Proper tests covering both preflight and runtime fallback paths
- Context preservation so users know WHY PyTorch was chosen
- Platform-aware messaging (Windows-specific guidance only on Windows)
- Bonus: pre-existing typo fix in pip install command

Minor non-blocking suggestions:
- Add trailing newline to `llm_backend_compat.py`
- Consider stripping the ❌ prefix from `vllm_fallback_note` before appending it to the ✅ success message

LGTM.

PR #844, opened 2026-03-16
fix: correct SFT model default inference steps from 32 to 50

PSiKoTiC:
get_ui_control_config() only distinguished between turbo (8 steps) and non-turbo (32 steps), with no detection for SFT models. This caused SFT models (acestep-v15-sft) to default to 32 inference steps on the GUI slider, despite their training defaults being optimized for 50 steps (as defined in training_v2/model_discovery._BASE_DEFAULTS).

Root cause: the turbo/non-turbo binary check was introduced during the UI decomposition refactor (36fe89ee) and carried over the generic 32-step fallback from the original gradio_ui/generation_handlers.py without accounting for the SFT model tier.

Fix:
- Add is_sft_model() detection function alongside existing is_pure_base_model(), checking for "sft" in the config path
- Thread is_sft parameter through get_ui_control_config(), get_model_type_ui_settings(), and update_model_type_settings()
- SFT models now correctly default to 50 steps; turbo remains at 8; pure base / unknown models retain the 32-step fallback
- Export is_sft_model from generation sub-package and facade module
- Correct tooltip description: SFT/Base default is 50 steps, not 100

No behavioral change for turbo or pure base models. Only SFT models are affected — their slider now initializes to the intended 50 steps.

## Summary by CodeRabbit

* **New Features**
  * Automatic detection and UI support for SFT (Supervised Fine-Tuned) models.

* **Improvements**
  * Updated inference-step defaults: SFT models default to 50 steps (30–100); Turbo unchanged; Base/unknown default to 32.
  * SFT status now consistently propagated through model UI configuration for more accurate defaults.

* **Tests**
  * Added unit tests for SFT detection and inference-step defaults.

* **Documentation**
  * Updated in-app guidance text to reflect SFT step ranges.

### DIFF
--- a/acestep/ui/gradio/events/generation/model_config.py
+++ b/acestep/ui/gradio/events/generation/model_config.py
@@ -5,6 +5,8 @@ producing UI control configurations, and computing gr.update() tuples
 for model-type-dependent controls.
 """
 
+import re
+
 import gradio as gr
 
 from acestep.constants import (
@@ -13,6 +15,15 @@ from acestep.constants import (
     GENERATION_MODES_TURBO,
     GENERATION_MODES_BASE,
 )
+
+
+def _has_token(token: str, path: str) -> bool:
+    """Check if *token* appears as a delimited word in *path*.
+
+    Matches when *token* is bounded by start/end of string or a common
+    path delimiter (``/``, ``\\``, ``.``, ``_``, ``-``).
+    """
+    return re.search(rf"(^|[\\\\/._-]){token}($|[\\\\/._-])", path) is not None
 
 
 def is_pure_base_model(config_path_lower: str) -> bool:
@@ -25,9 +36,9 @@ def is_pure_base_model(config_path_lower: str) -> bool:
         ``True`` when the path contains ``"base"`` and excludes ``"sft"`` and ``"turbo"``.
     """
     return (
-        "base" in config_path_lower
-        and "sft" not in config_path_lower
-        and "turbo" not in config_path_lower
+        _has_token("base", config_path_lower)
+        and not _has_token("sft", config_path_lower)
+        and not _has_token("turbo", config_path_lower)
     )
 
 
@@ -49,7 +60,7 @@ def update_model_type_settings(config_path: str | None, current_mode: str | None
 
     # Precedence: turbo > SFT > pure base > fallback.
     # Detection functions enforce mutual exclusivity.
-    is_turbo = "turbo" in config_path_lower
+    is_turbo = _has_token("turbo", config_path_lower)
     is_pure_base = is_pure_base_model(config_path_lower)
     is_sft = is_sft_model(config_path_lower)
 
@@ -65,7 +76,7 @@ def is_sft_model(config_path_lower: str) -> bool:
     Returns:
         ``True`` when the path contains ``"sft"`` and excludes ``"turbo"``.
     """
-    return "sft" in config_path_lower and "turbo" not in config_path_lower
+    return _has_token("sft", config_path_lower) and not _has_token("turbo", config_path_lower)
 
 
 def get_ui_control_config(is_turbo: bool, is_pure_base: bool = False, is_sft: bool = False) -> dict:
--- a/acestep/ui/gradio/events/generation/model_config_test.py
+++ b/acestep/ui/gradio/events/generation/model_config_test.py
@@ -3,11 +3,45 @@
 import unittest
 
 from acestep.ui.gradio.events.generation.model_config import (
+    _has_token,
     is_sft_model,
     is_pure_base_model,
     get_ui_control_config,
     update_model_type_settings,
 )
+
+
+class HasTokenTests(unittest.TestCase):
+    """Verify _has_token matches tokens at various delimiter boundaries."""
+
+    def test_token_with_hyphens(self):
+        """Standard hyphen-delimited path."""
+        self.assertTrue(_has_token("sft", "acestep-sft-1b"))
+
+    def test_token_at_start(self):
+        """Token at the beginning of the string."""
+        self.assertTrue(_has_token("sft", "sft-model"))
+
+    def test_token_at_end(self):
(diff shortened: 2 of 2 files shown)


ChuxiJ (changes requested):
Logic is correct — SFT 50-step default is well-supported by `model_discovery._BASE_DEFAULTS` and training CLI help text. The `is_sft_model()` detection and `get_ui_control_config()` update are sound.

**Blocking: missing unit tests** (per AGENTS.md, every behavior change must include tests). Please add:

1. Tests for `is_sft_model()` — positive (`"acestep-v15-sft"`), negative (`"acestep-v15-turbo"`, `"acestep-v15-base"`) inputs
2. Test `get_ui_control_config(is_turbo=False, is_sft=True)` returns 50 steps
3. Regression test: `get_ui_control_config(is_turbo=False, is_sft=False)` still returns 32 steps

Non-blocking: `zh.json` / `ja.json` / `he.json` tooltips don't mention SFT yet — acceptable if treated as best-effort.

ChuxiJ:
@copilot, please help update the i18n json

ChuxiJ (approved):
## Code Review

**Verdict: Approve ✅**

Clean, well-scoped fix. SFT models were incorrectly defaulting to 32 inference steps when their training config specifies 50. The `is_sft_model()` detection mirrors the existing `is_pure_base_model()` pattern, and the flag is correctly threaded through all three UI config functions.

### Positives
- `is_sft_model()` correctly excludes turbo models from SFT classification
- Service-mode initialization path is also updated (avoiding a subtle omission)
- Tests cover detection logic and step-value defaults for all three model types

### Minor suggestions (non-blocking)
- **i18n**: Only `en.json` is updated. The `zh.json`, `ja.json`, and `he.json` files also have `inference_steps_info` and still only mention "Turbo" and "Base" without SFT guidance. Non-English users will see stale tooltips.
- **Edge case**: The substring check `"sft" in config_path_lower` could false-positive on paths like `/users/sftuser/model/` — same limitation as `is_pure_base_model`, low risk in practice.
- Consider adding an integration test for `update_model_type_settings` with an SFT model path to verify the full chain.

Ready to merge as-is; i18n updates can be a follow-up.

ChuxiJ (commented):
## Code Review

**Overall: Clean, well-scoped bug fix. Suitable for merging with minor suggestions.**

### Suggestions

1. **Low - Substring matching fragility**: `is_sft_model` checks `"sft" in config_path_lower`, which would match substrings like "transform", "offset", etc. This is a pre-existing pattern (same as `is_pure_base_model` with "base"), but consider adding a test for substring false positives (e.g., `"transform-1b"`) to document the limitation.

2. **Low - Python version compatibility**: `str | None` type annotation (PEP 604) requires Python 3.10+ at runtime. Verify this is compatible with the project's minimum Python version, or add `from __future__ import annotations`.

3. **Nit**: Consider adding a comment in `get_ui_control_config` documenting the precedence: turbo > SFT > base fallback. Currently if `is_turbo=True, is_sft=True` is passed, turbo silently wins.

### Good Practices
- Mirrors the existing `is_pure_base_model()` pattern
- Proper export chain through `__init__.py`
- `config_lower` avoids calling `.lower()` twice
- Tests cover turbo, SFT, and base tiers

LGTM with the minor suggestions above.

PSiKoTiC:
## Review feedback addressed (commit a4805d0)

**Type hints and documentation**
- Added `-> tuple` return type and `Returns:` docstring to `update_model_type_settings`
- Added concise precedence comments at both decision points (`update_model_type_settings` and `get_ui_control_config`)

**Substring false-positive tests (reviewer suggestion)**
- `is_sft_model`: "sftp-server" contains "sft" → falsely detected as SFT
- `is_pure_base_model`: "database" contains "base" → falsely detected as base
- Integration test: proves the false positive flows through `update_model_type_settings` end-to-end (returns 50 steps instead of 32)
- Both detection functions use simple `in` checks — fixing requires word-boundary matching across all callers, which is a separate refactor. Tests document the current behavior so the limitation is visible

**Service-mode SFT fix (found during audit)**
- `service_init.py` was not passing `is_sft` to `get_model_type_ui_settings`, so SFT models initialized via service mode silently got 32 steps instead of the intended 50
- The interactive path (`generation_advanced_settings.py`) already passed `is_sft` correctly — this aligns service mode to match

### Design notes

- Precedence comments state "detection functions enforce mutual exclusivity" — this is accurate today because `is_sft_model` excludes "turbo" paths and `is_pure_base_model` excludes both. However, `get_ui_control_config` does not structurally enforce precedence — if multiple flags were somehow True, turbo would win for steps/visibility but pure-base would leak into task/mode choices. Safe as long as detection functions maintain their current exclusion logic, but worth noting if those functions are ever refactored
- `service_init.py` is 293 LOC (pre-existing, above the 200 LOC soft cap). Our change adds 4 net lines — flagging for awareness, not addressing here as it would be scope creep

### Validation

- 19/19 tests pass
- Two-pass defensive audit: clean (no warnings, no critical issues)
- No platform, hardware, or OS-specific assumptions in changes
- All existing callers unaffected (`is_sft` defaults to `False`)

PSiKoTiC:
## Word-boundary matching fix (commit b8e00f4)

Addresses the substring false-positive flagged by CodeRabbit.

**What changed:**
- Added `_has_token(token, path)` helper — regex-based delimiter-aware matching (`/`, `\`, `.`, `_`, `-`, start/end of string)
- `is_sft_model` and `is_pure_base_model` now use `_has_token` for all checks (detection and exclusion), applied consistently
- `update_model_type_settings` also updated for full consistency

**Scope note:** Previously stated this was a separate refactor. Applying here instead to avoid a merge-order dependency — `is_sft_model` only exists on this branch, so a standalone PR would depend on this one merging first. The scope expansion is noted for transparency.

**Validation:**
- 26/26 tests pass (7 new `_has_token` boundary tests, 3 false-positive tests flipped from documenting the bug to proving the fix)
- Three audit rounds: clean

PR #848, opened 2026-03-16
fix(nanovllm): make torch.compile conditional on Triton availability

PSiKoTiC:
## Summary

- Replaced all five `@torch.compile` decorators in vendored nanovllm with a conditional `@maybe_compile` that checks for Triton availability before compiling
- Unlocks the full nanovllm/vLLM inference path on Blackwell GPUs (RTX 5090/5080/5070) and any system without Triton installed

## Problem

On systems where Triton is unavailable — most commonly Windows, and NVIDIA Blackwell GPUs (SM 120) where Triton lacks architecture support — nanovllm crashes on the first LLM forward pass:

Five `@torch.compile` decorators across four layer files unconditionally require Triton as the compilation backend. When Triton is absent, the entire vLLM inference path fails and falls back to the significantly slower PyTorch backend.

The attention module (`attention.py`) already handles this correctly with a `_HAS_TRITON` flag and PyTorch fallback — the other layers simply lacked the same pattern.

## Fix

Added `nanovllm/utils/compat.py` with a `maybe_compile()` decorator:
- When Triton is importable: identical to `@torch.compile`
- When Triton is absent: returns the original function unmodified

Replaced `@torch.compile` in:
- `layers/activation.py` — `SiluAndMul.forward()`
- `layers/layernorm.py` — `RMSNorm.rms_forward()`, `RMSNorm.add_rms_forward()`
- `layers/rotary_embedding.py` — `RotaryEmbedding.forward()`
- `layers/sampler.py` — `Sampler.forward()`

## Platform impact

| Platform | Effect |
|----------|--------|
| Linux + Triton | Zero change — `torch.compile` still applied |
| Windows / no Triton | vLLM path now works (was crashing) |
| Blackwell (SM 120) | Full vLLM path unlocked |
| ROCm (AMD) | Unaffected (has own Triton) |
| MPS (Apple) | Unaffected (uses MLX backend) |

## Validation

Tested on RTX 5090 (SM 120, 32 GB VRAM, Windows 11 Pro 24H2, no Triton):

- All four layer modules (SiluAndMul, RMSNorm, RotaryEmbedding, Sampler) import and execute on GPU without errors
- `flash_attn` 2.7.4 confirmed working independently on SM 120
- CUDA graphs confirmed working on SM 120
- Native PyTorch kernels provide correct output (verified against expected tensor shapes)

## Test plan

- [ ] Verify vLLM backend initializes successfully on a Blackwell GPU (RTX 50-series)
- [ ] Verify existing behavior unchanged on systems with Triton installed
- [ ] Run LLM text generation end-to-end via nanovllm on SM 120
- [ ] Confirm no performance regression on Triton-equipped systems (torch.compile still applied)

## Summary by CodeRabbit

* **Refactor**
  * Switched compilation decorators to a compatibility-aware variant across inference and model layers to allow optional compilation while preserving behavior.

* **New Features**
  * Added runtime detection of Triton availability and a safe fallback so compilation is skipped when Triton/PyTorch features are missing.

* **Bug Fixes / Stability**
  * Disable CUDA graph capture and suppress verbose Dynamo warnings when Triton is absent; startup logs now explain reduced-capability fallbacks.

### DIFF
--- a/acestep/llm_inference.py
+++ b/acestep/llm_inference.py
@@ -600,11 +600,12 @@ class LLMHandler:
                     "(SDPA fallback uses .item() calls in paged-cache decode that are "
                     "incompatible with CUDA graph capture)"
                 )
+            _has_triton = False
             try:
                 import triton  # noqa: F401
                 _has_triton = True
             except ImportError:
-                _has_triton = False
+                pass
             if not _has_triton:
                 logger.info(
                     "Triton not available: disabling CUDA graph capture for nano-vllm "
--- a/acestep/third_parts/nano-vllm/nanovllm/utils/compat.py
+++ b/acestep/third_parts/nano-vllm/nanovllm/utils/compat.py
@@ -11,6 +11,15 @@ from loguru import logger
 
 F = TypeVar("F", bound=Callable[..., Any])
 
+# Module-level Triton availability check — runs once at import time
+# rather than repeating the import probe at every decoration site.
+_HAS_TRITON = False
+try:
+    import triton  # noqa: F401
+    _HAS_TRITON = True
+except ImportError:
+    pass
+
 
 def maybe_compile(fn: Optional[F] = None, **compile_kwargs: Any) -> Any:
     """Apply ``torch.compile`` only when its backend (Triton) is available.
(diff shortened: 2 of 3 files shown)


ChuxiJ (approved):
Re-reviewed after the fix commit. All three requested changes are properly addressed:

1. **Unit tests added** — `compat_test.py` (45 LOC) covers all 3 scenarios: Triton available (compiled), Triton absent (original function returned), kwargs syntax (`dynamic=True`) forwarded. Deterministic, fast, properly mocked.

2. **Defensive default** — `_has_triton = False` initialized before `try` block in `llm_inference.py`.

3. **Module-level constant** — `_HAS_TRITON` in `compat.py` now runs once at import time instead of per-decoration-site.

The `try`/`finally` dynamo state restoration in `_initialize_5hz_lm_vllm` is correctly structured. The `_dynamo_state_saved` flag properly gates the `finally` cleanup.

LGTM.

PR #852, opened 2026-03-16
Selectable MP3 export format

robustini:
Replaces the implicit MP3 export path in torchaudio/ffmpeg by adding the option to select the MP3 format.
If the UI does not support these settings, generation will default to the previous settings.

## Release Notes

* **New Features**
  * Added MP3 export support with customizable bitrate and sample rate configuration
  * Enabled batch audio file processing capabilities
  * Introduced audio format conversion utility for easy format switching
  * Added audio processing effects including fade and normalization

### DIFF
--- a/acestep/core/generation/handler/generate_music.py
+++ b/acestep/core/generation/handler/generate_music.py
@@ -54,6 +54,53 @@ class GenerateMusicMixin:
     orchestration flow.
     """
 
+    def _verify_decoder_device_dtype(self) -> None:
+        """Ensure all decoder parameters are on ``self.device`` and ``self.dtype``.
+
+        CPU-offload round-trips can leave PEFT/LoRA adapter weights on the
+        wrong device or in the wrong dtype, which silently produces NaN/Inf
+        during the diffusion forward pass.  This method detects and fixes
+        such mismatches before generation begins.
+        """
+        decoder = getattr(self.model, "decoder", None)
+        if decoder is None:
+            return
+
+        wrong_device = []
+        wrong_dtype = []
+        for name, param in decoder.named_parameters():
+            if not self._is_on_target_device(param, self.device):
+                wrong_device.append(name)
+            if param.is_floating_point() and param.dtype != self.dtype:
+                wrong_dtype.append(name)
+
+        if wrong_device or wrong_dtype:
+            if wrong_device:
+                logger.warning(
+                    f"[generate_music] LoRA sanity check: {len(wrong_device)} decoder "
+                    f"parameters on wrong device (expected {self.device}), fixing: "
+                    f"{wrong_device[:5]}{'...' if len(wrong_device) > 5 else ''}"
+                )
+            if wrong_dtype:
+                logger.warning(
+                    f"[generate_music] LoRA sanity check: {len(wrong_dtype)} decoder "
+                    f"parameters have wrong dtype (expected {self.dtype}), fixing: "
+                    f"{wrong_dtype[:5]}{'...' if len(wrong_dtype) > 5 else ''}"
+                )
+            decoder.to(device=self.device, dtype=self.dtype)
+
+        # Final verification — use recursive move if simple .to() wasn't enough
+        still_wrong = [
+            name for name, p in decoder.named_parameters()
+            if not self._is_on_target_device(p, self.device)
+        ]
+        if still_wrong:
+            logger.warning(
+                f"[generate_music] {len(still_wrong)} params still on wrong device "
+                f"after decoder.to(), using recursive move"
+            )
+            self._recursive_to_device(decoder, self.device, self.dtype)
+
     def _vram_preflight_check(
         self,
         actual_batch_size: int,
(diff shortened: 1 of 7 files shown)


ChuxiJ (changes requested):
There are still some other issues with 44.1kHz.
We still recommend 48kHz, and suggest making it **user-selectable** here instead of hardcoding it.

robustini:
> There are still some other issues with 44.1kHz. We still recommend 48kHz, and suggest making it **user-selectable** here instead of hardcoding it.

That makes sense.
In the original implementation, MP3 export went through torchaudio.save(..., backend="ffmpeg"), but bitrate and MP3-specific ffmpeg options were not actually exposed/passed through there, so there was no real user-selectable path for kbps/sample-rate tuning.
My change introduced an explicit MP3 export path so those parameters can be controlled deterministically. I agree that 48 kHz should remain the default/recommended option, and 44.1 kHz should be exposed as a user-selectable export setting rather than hardcoded.
I can revise the PR in that direction.

robustini:
@ChuxiJ the export path now honors explicit MP3 bitrate and sample-rate values when they are provided by the UI.
If those parameters are not sent, MP3 export falls back to the existing default behavior exactly as before.
I haven't touched the Gradio, do it yourself if you think it's appropriate, tested with AceFlow and it works.

immagine

ChuxiJ (changes requested):
## Code Review

**Verdict: Needs Changes ❌**

The MP3 export feature itself is well-implemented (`_save_mp3()` with proper temp file cleanup, input validation, timeout), but the `inference.py` refactoring introduces a functional regression.

### Critical: Dropped `global_caption` parameter

The refactoring converts the `dit_handler.generate_music()` call from explicit keywords to a dict + `inspect.signature()` filter. However, the new dict **omits** `global_caption=params.global_caption` which is present in the current main branch and accepted by the handler. The user's `global_caption` value will be silently ignored — this is a **functional regression**.

`chunk_mask_mode` is also missing from the dict, though its default `"auto"` matches the previous `getattr` fallback, making it benign.

### Medium: `inspect.signature()` pattern is fragile

This pattern silently drops parameters without error, making bugs like the above easy to introduce. A misspelled key would be silently filtered out instead of raising `TypeError`. The `logger.debug` for dropped keys is easy to miss. The previous explicit-keyword approach was safer.

### Required changes
1. Add `global_caption` and `chunk_mask_mode` back to `dit_generate_kwargs`
2. Consider raising a warning (not just debug) when parameters are dropped by the signature filter

### Other suggestions (non-blocking)
- Forward `mp3_bitrate`/`mp3_sample_rate` in `convert_audio()`, `save_batch()`, and the module-level `save_audio()` for consistency
- Change `logger.info` at the export log line to `logger.debug` to reduce noise in batch generation
- Minimize the `audio_utils.py` diff to only actual semantic changes (current diff is a full-file replacement)

ChuxiJ (commented):
## Code Review

**Overall: Well-implemented MP3 export with proper error handling, input validation, and test coverage.**

### Key Items to Address

1. **Medium - API gap for `convert_audio` and `save_batch`**: These methods (and the module-level `save_audio` convenience function) do not accept the new `mp3_bitrate`/`mp3_sample_rate` parameters. Users calling these functions cannot control MP3 export settings. At minimum, thread the parameters through `convert_audio` and `save_batch`, or document this limitation.

2. **Medium - Whole-file rewrite inflates diff**: The diff shows -516/+596 lines for `audio_utils.py`, but the actual changes are ~60 lines. This was likely caused by re-formatting. A cleaner diff would make review easier.

3. **Low - `logger.info` in hot path**: The new `logger.info` at line 743 in `inference.py` logs on every audio export. For batch generation this is noisy. Consider using `logger.debug` to match the existing logging style in `save_audio`.

4. **Low - Test formatting**: Multi-`patch` decorators on single lines are hard to read. Consider stacking or using `with` blocks.

### Good Practices
- `subprocess.run` with list args prevents shell injection
- Input validation on bitrate and sample rate with sensible defaults
- `timeout=120` on subprocess prevents runaway ffmpeg processes
- Temp file cleanup in `finally` block
- `inspect.signature` filtering in `inference.py` is a clever forward-compatibility mechanism

No blocking issues.

ChuxiJ:
### Code review

The core `_save_mp3` mechanism (explicit ffmpeg subprocess with temp WAV) is sound, and test coverage is solid.

Found 2 issues:

1. **`repaint_*` fields in `GenerationParams` are dead code** -- `repaint_latent_crossfade_frames`, `repaint_wav_crossfade_sec`, `repaint_mode`, and `repaint_strength` are declared in `GenerationParams` (lines 144-147) but never added to `dit_generate_kwargs` (lines 593-628). The `inspect.signature` filtering shim cannot pass keys that are not in the dict. Any non-default values set by callers will be silently dropped. These four keys need to be added to `dit_generate_kwargs`.

2. **Default bitrate (128k) contradicts stated purpose (320k)** -- The PR title, branch name (`mp3-320k-44k1`), and early commit messages claim the goal is 320 kbps CBR. But `MP3_DEFAULT_BITRATE = "128k"` and `mp3_bitrate: str = "128k"` mean every user gets 128k unless they explicitly override. No UI wiring for `mp3_bitrate`/`mp3_sample_rate` is included in this PR, so the user-facing behavior is unchanged from before. Either update the description or raise the default.

Minor notes:
- `getattr(config, "mp3_bitrate", "128k")` is redundant since both are proper dataclass fields
- ffmpeg command uses ABR mode by default (not CBR as stated) -- for true CBR, needs `-abr 0`
- Unused imports: `io`, `Tuple` in `audio_utils.py`

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #854, opened 2026-03-17
Bugs after updating during generation


ChuxiJ:
### Discussed in https://github.com/ace-step/ACE-Step-1.5/discussions/850

<div type='discussions-op-text'>

<sup>Originally posted by **tornado73** March 16, 2026</sup>
Hello, yesterday I updated Ace Step and here's what I noticed.
I'm generating on Acestep-v15-base + Acestep-5Hz-lm-1.7B, with the Gradio interface. After 5-10 generations, it starts overlaying random sounds onto the composition. The intro is completely unpredictable and plays nonsense 99% of the time. I've tried disabling the intro in various ways, but it still keeps popping it everywhere. Sometimes reinitializing the models helps, but it doesn't last long. After 1-2 generations, it freezes dead at the stage. 2026-03-16 11:53:57.952 | WARNING  | acestep.core.generation.handler.vae_decode_chunks:_tiled_decode_inner:35 - [tiled_decode] Reduced overlap from 64 to 32 for chunk_size=128, , This didn't happen in the previous version.
A clean install didn't help, it's still the same.
start data

Model: D:\ace\ACE-Step-1.5\checkpoints\acestep-5Hz-lm-1.7B
Device: NVIDIA GeForce RTX 4070 Ti SUPER
GPU Memory Utilization: 0.527
Low GPU Memory Mode: False
5Hz LM initialized successfully
Service initialization completed successfully!
Creating Gradio interface with language: en...
Enabling queue for multi-user support...

p.s.  Acestep-v15-sft    works fine
</div>

FocusMode-coder:
Hey! I may be able to help with this bug.

I build custom fixes for Python projects and automation systems.

If you'd like help resolving this issue quickly feel free to reach out.

Happy to take a look 👍

ChuxiJ:
Thanks for the detailed report. This sounds like a state/cache leak in the generation pipeline that accumulates over successive runs.

The key clues:
- **SFT model works fine** — this suggests the leak is in the LM code path (which SFT bypasses)
- **tiled_decode overlap reduction warning** — may indicate buffer sizes getting corrupted in the VAE decode path
- **Progressive degradation** — characteristic of mutable state not being properly cleared between generations

We're investigating the following areas:
1. VAE tiled decode path (`vae_decode_chunks.py`) — checking if overlap buffers persist across calls
2. LM code tensors and reference audio caches — checking for references that leak between generation calls
3. Mutable class-level state in the generation handler

[...]

PR #869, opened 2026-03-18
#854: Restore GPU synchronization to prevent VRAM race conditions (audio overlays/deadlocks)

Uni404x64:
The refactor in 9ea78c1 (renamed _clear_cuda_cache to _clear_accelerator_cache) dropped the critical torch.cuda.synchronize() call. Without it, empty_cache() releases memory while the GPU still has async writes in flight, causing cross-generation memory races (audio overlays, prompt drift, allocation deadlocks).

- Restore synchronize() after empty_cache() on CUDA path
- Add synchronize() on XPU and MPS paths (hasattr-guarded)
- Update docstring to document the sync barrier and its purpose
- Update tests to assert synchronize() is called on all backends

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure device operations are synchronized before clearing accelerator caches on CUDA, XPU, and MPS to prevent cross-generation data races and improve memory cleanup; includes a safe fallback when device selection is unset.

* **Tests**
  * Updated test expectations to verify synchronization occurs before cache clearing across all supported backends and the fallback scenario.

### DIFF
--- a/acestep/llm_inference_cache_cleanup_test.py
+++ b/acestep/llm_inference_cache_cleanup_test.py
@@ -62,4 +62,21 @@ class LlmAcceleratorCacheCleanupTests(unittest.TestCase):
         """Method should be a safe no-op when device is CPU."""
         handler = LLMHandler()
         handler.device = "cpu"
-        with patch
+        with patch("torch.cuda.empty_cache") as cuda_mock:
+            handler._clear_accelerator_cache()
+        cuda_mock.assert_not_called()
+
+    def test_falls_back_to_cuda_when_device_unset(self):
+        """When self.device is None, should fall back to CUDA if available."""
+        handler = LLMHandler()
+        handler.device = None
+        call_order = []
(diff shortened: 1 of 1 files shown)


Uni404x64:
@ChuxiJ — appreciate the review. You're right that the diff looks insane, but that's not what's actually happening here.

My fork fell behind main, and the PR diff is comparing my branch (which has the old version of llm_inference.py + the sync fix) against current main (which has had a bunch of changes since I forked). So GitHub is showing the delta between those two diverged versions as if I'm rewriting the whole file. I'm not.

The actual change is exactly what the description says — restoring synchronize() before empty_cache() on the CUDA/XPU/MPS paths, plus the test updates. It didn't display like this when I opened it.

I also had a browser editor issue that truncated the test file on one commit (c686bd4 fixes that), which didn't help.

I'm going to sync my fork with upstream and rebase so the diff shows only the actual changes. Give me a bit to get that sorted and the PR should look clean.

PR #888, opened 2026-03-20
fix(i18n): add per-request language isolation via ContextVar

### LINKED ISSUE #880: i18n: global language state is request-unsafe under concurrent access
The `I18n` singleton in `i18n/i18n.py` stores `current_language` as a mutable instance attribute. When `get_i18n(language)` or `set_language()` is called, it mutates this shared state. Under concurrent Gradio requests (e.g., two users with different language preferences), one request can overwrite `current_language` while another is mid-translation via `t()`.

The lock added in #878 serializes writes but doesn't prevent cross-request leakage — request A sets language to "zh", request B sets it to "en" before A finishes translating, and A gets English strings.

A clean fix would be moving `current_language` out of the singleton and into a `contextvars.ContextVar`, so each async request context carries its own language. The `I18n` instance stays shared (translations are read-only dicts, safe to share), but language selection becomes request-local.

This would only touch `i18n.py` — the `t()` and `get_i18n()` API can stay identical, so the 47+ files calling `t("key")` wouldn't need changes.

Identified during review of #878.

PSiKoTiC:
## Summary

- Added `contextvars.ContextVar`-based per-request language scoping to `I18n.t()`, so concurrent requests with different language preferences no longer interfere with each other.
- Builds on the thread-safe locking from #878 — all existing locks are preserved, ContextVar is an additional layer.
- Provides `set_language_context()` / `reset_language_context()` public API for request-boundary scoping.

## Problem

The `I18n` singleton stores `current_language` as a mutable instance attribute. When `set_language()` is called from one Gradio request, it mutates shared state visible to all concurrent requests. The locks added in #878 serialise writes but don't prevent cross-request leakage — request A sets language to "zh", request B sets it to "en" before A finishes translating, and A gets English strings.

Identified in Discussion #870 and tracked in #880.

## Fix

- **`i18n.py`**: Added `_current_language_var: ContextVar[str | None]` with `default=None`. `I18n.t()` checks the ContextVar first; when unset (`None`), falls back to the locked `self.current_language` — preserving identical behaviour for all existing callers. Added `set_language_context()` (with `loguru` warning on invalid language codes) and `reset_language_context()` as the public API. The ContextVar is set *before* any validation/logging to prevent reentrant `t()` calls from seeing stale state.
- **`__init__.py`**: Exports `set_language_context` and `reset_language_context`.
- **`generation_service_wiring.py`**: `_apply_runtime_language` now wraps the language switch in `try/finally` with token cleanup, preventing thread-pool worker poisoning from stale ContextVar values. Also updates the shared instance default via `get_i18n(language)` so future requests without an explicit context inherit the last-selected language.
- **`i18n_thread_safety_test.py`**: 7 new tests (9 total): end-to-end translation output verification across languages, ContextVar fallback to instance default, nested token reset, concurrent thread isolation with barrier sync, mixed context/fallback thread scenario, invalid language warning coverage, and `copy_context` isolation (validates async framework compatibility).

No function signatures, return values, or error paths changed for existing callers. The 47+ files importing `t()` are completely unaffected.

## Module size note

`i18n.py` is 218 LOC (project guideline: ≤200). The pre-existing file was 195 LOC before this change. The 23-line increase covers two new public functions (`set_language_context`, `reset_language_context`) with mandatory docstrings. Splitting into a separate module was evaluated but rejected — the ContextVar, its setter/resetter, and the `I18n.t()` method that reads it are tightly coupled, and extracting them would require either circular imports or an indirection layer that adds complexity without improving readability.

## Not addressed (pre-existing)

Pre-existing style items in `i18n.py` were identified during the audit but are out of scope for this PR. Documented in [Discussion #870 comment](https://github.com/ace-step/ACE-Step-1.5/discussions/870#discussioncomment-16225337) for future cleanup.

## Platform impact

- Pure Python `contextvars` (stdlib since 3.7), no hardware dependencies
- `str | None` syntax requires Python ≥3.10 (project requires ≥3.11)
- Works identically on Windows, Linux, macOS
- No CUDA/MPS/XPU/CPU code paths affected

## Test plan

- 9 unit tests pass locally (2 existing singleton safety + 7 new ContextVar isolation)
- End-to-end: `t("app.title")` returns different translations when ContextVar switches between en/zh
- Concurrent isolation: two threads with different ContextVar values don't interfere
- Mixed scenario: one thread with ContextVar override, one without, each gets correct language
- Token cleanup: `reset_language_context` restores prior value through nested set/reset
- Invalid language: warning logged, graceful English fallback
- `copy_context` isolation: child context changes don't leak to parent (async framework compat)

Closes #880

## Summary by CodeRabbit

* **New Features**
  * Per-request language context isolation to ensure concurrent requests can use different language settings without conflicts.
  * Now emits a warning when an unsupported language is requested, improving diagnostics.

* **Tests**
  * Added comprehensive tests verifying per-request language isolation, thread/context safety, fallback behavior, and correct restoration of prior language state.

### DIFF
--- a/acestep/ui/gradio/i18n/i18n_thread_safety_test.py
+++ b/acestep/ui/gradio/i18n/i18n_thread_safety_test.py
@@ -1,8 +1,10 @@
 """Thread safety tests for the i18n singleton and ContextVar isolation."""
 
 import contextvars
+import sys
 import threading
 import unittest
+from unittest.mock import patch
 
 from acestep.ui.gradio.i18n.i18n import (
     _current_language_var,
@@ -188,17 +190,22 @@ class ContextVarIsolationTests(unittest.TestCase):
         # Ensure singleton exists so validation can run
         get_i18n()
         # "xyz_invalid" is not a loaded translation
-        token = set_language_context("xyz_invalid")
-        try:
-            # ContextVar is still set (graceful degradation)
-            self.assertEqual(_current_language_var.get(), "xyz_invalid")
-            # t() should fall back to English, then to key itself
-            result = t("app.title")
-            # Should get the English fallback, not the raw key
-            self.assertNotEqual(result, "app.title",
-                                "Expected English fallback, got raw key")
-        finally:
-            reset_language_context(token)
+        with patch.object(
+            sys.modules[set_language_context.__module__], "logger",
+        ) as mock_logger:
+            token = set_language_context("xyz_invalid")
+            try:
+                # ContextVar is still set (graceful degradation)
+                self.assertEqual(_current_language_var.get(), "xyz_invalid")
+                # Warning must have fired exactly once
+                mock_logger.warning.assert_called_once()
+                # t() should fall back to English, then to key itself
+                result = t("app.title")
+                # Should get the English fallback, not the raw key
+                self.assertNotEqual(result, "app.title",
+                                    "Expected English fallback, got raw key")
+            finally:
+                reset_language_context(token)
 
     def test_contextvar_isolation_in_copy_context(self):
         """contextvars.copy_context isolates changes from the parent."""


PSiKoTiC:
Addressed the two nitpicks:

**`i18n_thread_safety_test.py` — token cleanup consistency:** Fixed in cdd725b. `test_contextvar_isolation_in_copy_context` now captures the token and resets via `try/finally`, matching every other test in the suite.

**`i18n.py` — pre-existing `print()` calls:** Those `print()` statements at lines 56 and 72 predate this PR. Already flagged in [Discussion #870](https://github.com/ace-step/ACE-Step-1.5/discussions/870#discussioncomment-16225337) as part of a broader style cleanup list. Keeping this PR scoped to the ContextVar work.

PSiKoTiC:
Addressed the logger assertion — `test_set_language_context_warns_on_invalid_language` now patches `logger` and asserts `warning` is called exactly once when an unknown language code is passed. Fixed in acf7dd9.

On the other two:

**Cross-locale inequality oracle (lines 71-74, 183-184):** Intentional. Hardcoding expected translations (e.g., `assertEqual(zh_title, "某个中文字符串")`) couples the test to specific translation file content — if a translator updates the string, the test breaks even though the ContextVar plumbing is fine. The `assertNotEqual` checks that the mechanism switched languages, without depending on what the translations say. `app.title` is "ACE-Step" in English and Chinese characters in zh — these won't converge.

**Non-English fallback test (lines 82-90):** The test's purpose is verifying the fallback *path* (ContextVar → instance default → English → key itself) completes without error when ContextVar is `None`. Using a nonexistent key exercises the full chain. Switching to a real locale would test translation content, not the fallback mechanism.

Issue #142, opened 2026-02-05
fills up RAM after 2 generations leading to PC lockup


screwballl:
**Describe the bug**
Every generation reserves system RAM, then does not release it after finishing generation. With 32GB RAM it fills up and hard locks the PC by the 3rd or 4th gen attempt. Tested with 0.6B and 1.7B LM model. First gen increases system RAM usage from 1.5GB to 13.3GB, 2nd generation jumps to 16.5 to 21.4GB usage, 3rd generation jumps to 30 to 32GB+.
Almost every time after 2nd gen the default GUI in the browser keeps the prior gens in the results section instead of adding the new gens.

**Expected behavior**
Code should clear/free RAM at least partially and reuse the RAM loaded processes instead of starting new ones every generation. 

**OS and hardware**
Linux Mint 22.3, Ryzen 9600X, 32GB DDR5-6000, 16GB RTX5060Ti

**context**
closing and restarting it, initialize service, and refresh page sometimes allows up to 4th gen but it is rare. Close and restart and refresh GUI in browser clears settings and lyrics entered, and language chosen.

imsarang:
Hi is this issue still there, can take this up!
@jaretburkett @saltchicken @fspecii @xmhu @screwballl

screwballl:
Updated with latest files, issue persists

imsarang:
ok, gonna take this up!
@screwballl @jaretburkett @saltchicken @fspecii @xmhu

rkfg:
This is indeed an issue. I run the Gradio UI with the following parameters: `--server-name 0.0.0.0 --service_mode true --offload_to_cpu true --offload_dit_to_cpu true --enable-api`
The `.env` file contains this:
```
ACESTEP_CONFIG_PATH=acestep-v15-sft
ACESTEP_LM_MODEL_PATH=acestep-5Hz-lm-4B
```
Each generation adds about 8 GB to the python process (the DiT size?). Even though I have 128 GB of RAM it fills up quickly because I also run other memory hungry programs like browser. I tried to exclude the LM as much as possible to narrow the range of possible culprits. Turned off CoT options, used a pre-generated audio codes, disabled "Think" of course. My workflow is regenerating the same batch of two songs with two fixed seeds and changing the lyrics a little between the runs. The log shows a lot of offloading of VAE, text encoder, and main model.

[...]

PR #947, opened 2026-03-27
fix(memory): eliminate system RAM leak in CPU offload mode (#142)

ChuxiJ:
## Summary

Fixes the system RAM fill-up issue (#142) where users running `--offload_to_cpu` or `--offload_dit_to_cpu` see ~4-8 GB of RAM leaked per generation, filling 32 GB RAM in 3-4 generations and hard-locking the PC.

**Root causes fixed (5-agent analysis):**

- **Redundant parameter sweeps**: `_recursive_to_device` called `_move_module_recursive` unconditionally after `model.to()` succeeded — 3× redundant allocations per transfer
- **Unsafe `dir()+getattr()` loop**: triggered computed properties and created transient objects on every submodule
- **New `nn.Parameter()` wrappers** in fallback path: multiplied temporary allocations instead of using in-place `param.data =` replacement
- **Dual `.to()` calls**: separate `model.to(device)` + `model.to(dtype)` doubled peak temporary storage; merged into single `model.to(device=..., dtype=...)`
- **No OS-level memory reclaim**: `gc.collect()` + `empty_cache()` insufficient; added `malloc_trim(0)` on Linux to return freed heap pages to OS
- **Peak RSS monitoring**: `_get_rss_mb()` used `ru_maxrss` (peak, never decreases) — fixed to use `/proc/self/statm` on Linux for current RSS
- **`M_MMAP_THRESHOLD` not set**: large allocs stayed in ptmalloc2 arena; now set to 64 KB at import via `mallopt(-3, 65536)` so freed memory returns to OS immediately
- **`batch_queue` tensor accumulation**: Gradio UI kept `extra_outputs` (~4-8 GB intermediate tensors) for all historical batches indefinitely; now deleted after 2+ generations

## Files changed

| File | Change |
|------|--------|
| `init_service_memory_basic.py` | Add `_get_rss_mb()` (current RSS via `/proc/self/statm`), `_release_system_memory()`, set `M_MMAP_THRESHOLD=65536` at import |
| `init_service_memory_transfer.py` | Remove redundant passes, remove `dir()+getattr()` loop, in-place `param.data` replacement, single `.to()` call |
| `init_service_offload_context.py` | Use `_release_system_memory()` after load and offload; add RSS delta logging |
| `batch_queue.py` | Delete `extra_outputs` from batches 2+ generations old |
| `init_service_test.py` | Update assertion: `_empty_cache` now called twice (load + offload) |

## Test plan

- [x] All 59 `init_service_test.py` tests pass
- [x] 686 total tests pass; 17 pre-existing failures unchanged
- [ ] Manual verification: run server with `--offload_to_cpu`, generate 5+ songs, monitor RSS with `htop` — should remain stable instead of growing ~4-8 GB per generation

## Summary by CodeRabbit

* **New Features**
  * Added optional save-memory mode (env var) that reduces stored intermediates and updates UI/messages when enabled.
  * Added user-facing message translations indicating LRC is disabled in save-memory mode.

* **Refactor**
  * Improved model/device transfer logic and added Linux RSS monitoring and proactive memory release to reduce peak RAM.
  * Aggressive cleanup of older batch tensor data to free memory.

* **Tests**
  * Adjusted test expectations to reflect additional memory-release calls.

### DIFF
--- a/acestep/core/generation/handler/generate_music_payload.py
+++ b/acestep/core/generation/handler/generate_music_payload.py
@@ -2,6 +2,7 @@
 
 from typing import Any, Dict
 
+from acestep.gpu_config import get_global_gpu_config
 from loguru import logger
 
 
@@ -46,39 +47,50 @@ class GenerateMusicPayloadMixin:
         status_message = "Generation completed successfully!"
         logger.info(f"[generate_music] Done! Generated {len(audio_tensors)} audio tensors.")
 
-        src_latents = outputs.get("src_latents")
-        target_latents_input = outputs.get("target_latents_input")
-        chunk_masks = outputs.get("chunk_masks")
-        spans = outputs.get("spans", [])
-        latent_masks = outputs.get("latent_masks")
+        # In save-memory mode, skip storing intermediate tensors to reduce RAM
+        # usage (~4-8 GB per generation). Only non-tensor metadata is kept.
+        save_memory = get_global_gpu_config().save_memory_mode
 
-        encoder_hidden_states = outputs.get("encoder_hidden_states")
-        encoder_attention_mask = outputs.get("encoder_attention_mask")
-        context_latents = outputs.get("context_latents")
-        lyric_token_idss = outputs.get("lyric_token_idss")
+        if save_memory:
+            extra_outputs = {
+                "spans": outputs.get("spans", []),
+                "time_costs": time_costs,
+                "seed_value": seed_value_for_ui,
+            }
+        else:
+            src_latents = outputs.get("src_latents")
+            target_latents_input = outputs.get("target_latents_input")
+            chunk_masks = outputs.get("chunk_masks")
+            spans = outputs.get("spans", [])
+            latent_masks = outputs.get("latent_masks")
 
-        extra_outputs = {
-            "pred_latents": pred_latents_cpu,
-            "target_latents": target_latents_input.detach().cpu() if target_latents_input is not None else None,
-            "src_latents": src_latents.detach().cpu() if src_latents is not None else None,
-            "chunk_masks": chunk_masks.detach().cpu() if chunk_masks is not None else None,
-            "latent_masks": latent_masks.detach().cpu() if latent_masks is not None else None,
-            "spans": spans,
-            "time_costs": time_costs,
-            "seed_value": seed_value_for_ui,
-            "encoder_hidden_states": (
-                encoder_hidden_states.detach().cpu()
-                if encoder_hidden_states is not None
-                else None
-            ),
-            "encoder_attention_mask": (
-                encoder_attention_mask.detach().cpu()
-                if encoder_attention_mask is not None
-                else None
-            ),
-            "context_latents": context_latents.detach().cpu() if context_latents is not None else None,
-            "lyric_token_idss": lyric_token_idss.detach().cpu() if lyric_token_idss is not None else None,
-        }
+            encoder_hidden_states = outputs.get("encoder_hidden_states")
+            encoder_attention_mask = outputs.get("encoder_attention_mask")
+            context_latents = outputs.get("context_latents")
+            lyric_token_idss = outputs.get("lyric_token_idss")
+
+            extra_outputs = {
+                "pred_latents": pred_latents_cpu,
+                "target_latents": target_latents_input.detach().cpu() if target_latents_input is not None else None,
+                "src_latents": src_latents.detach().cpu() if src_latents is not None else None,
+                "chunk_masks": chunk_masks.detach().cpu() if chunk_masks is not None else None,
+                "latent_masks": latent_masks.detach().cpu() if latent_masks is not None else None,
+                "spans": spans,
+                "time_costs": time_costs,
+                "seed_value": seed_value_for_ui,
+                "encoder_hidden_states": (
+                    encoder_hidden_states.detach().cpu()
+                    if encoder_hidden_states is not None
+                    else None
+                ),
+                "encoder_attention_mask": (
+                    encoder_attention_mask.detach().cpu()
+                    if encoder_attention_mask is not None
+                    else None
+                ),
+                "context_latents": context_latents.detach().cpu() if context_latents is not None else None,
+                "lyric_token_idss": lyric_token_idss.detach().cpu() if lyric_token_idss is not None else None,
+            }
 
         audios = []
         for audio_tensor in audio_tensors:
--- a/acestep/gpu_config.py
+++ b/acestep/gpu_config.py
@@ -23,6 +23,7 @@ from loguru import logger
 DEBUG_MAX_CUDA_VRAM_ENV = "MAX_CUDA_VRAM"
 DEBUG_MAX_MPS_VRAM_ENV = "MAX_MPS_VRAM"
 DEBUG_MAX_XPU_VRAM_ENV = "MAX_XPU_VRAM"
+SAVE_MEMORY_ENV = "ACESTEP_SAVE_MEMORY"
 
 # Tolerance for 16GB detection: reported VRAM like 15.5GB is effectively 16GB hardware
 # Real-world 16GB GPUs often report 15.7-15.9GB due to system/driver reservations
@@ -214,6 +215,11 @@ class GPUConfig:
 
     # LM memory allocation (GB) for each model size
     lm_memory_gb: Dict[str, float]  # e.g., {"0.6B": 3, "1.7B": 8, "4B": 12}
+
+    # Save-memory mode: skip storing intermediate tensors in extra_outputs
+    # and disable auto_lrc / auto_score to reduce RAM usage.
+    # Controlled via ACESTEP_SAVE_MEMORY=1 environment variable.
+    save_memory_mode: bool = False
 
 
 def _apply_lm_backend_compatibility_overrides(config: GPUConfig) -> GPUConfig:
@@ -1464,10 +1470,18 @@ _global_gpu_config: Optional[GPUConfig] = None
 
 
 def get_global_gpu_config() -> GPUConfig:
-    """Get the global GPU configuration, initializing if necessary."""
+    """Get the global GPU configuration, initializing if necessary.
+
+    Respects the ``ACESTEP_SAVE_MEMORY`` environment variable: when set to
+    ``"1"`` or ``"true"``, ``save_memory_mode`` is enabled regardless of tier.
+    """
     global _global_gpu_config
     if _global_gpu_config is None:
         _global_gpu_config = get_gpu_config()
+        env_val = os.environ.get(SAVE_MEMORY_ENV, "").strip().lower()
+        if env_val in ("1", "true", "yes"):
+            _global_gpu_config.save_memory_mode = True
+            logger.info("[gpu_config] Save-memory mode enabled via {}={}".format(SAVE_MEMORY_ENV, env_val))
     return _global_gpu_config
 
 
--- a/acestep/ui/gradio/events/results/generation_progress.py
+++ b/acestep/ui/gradio/events/results/generation_progress.py
@@ -72,6 +72,11 @@ def generate_with_progress(
     gpu_config = get_global_gpu_config()
     lm_initialized = llm_handler.llm_initialized if llm_handler else False
 
+    # Save-memory mode: force-disable features that require intermediate tensors
+    if gpu_config.save_memory_mode:
+        auto_score = False
+        auto_lrc = False
+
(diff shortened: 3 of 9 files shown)


acestep/ui/gradio/events/results/batch_queue.py
```diff
@@ -54,6 +54,12 @@ def store_batch_in_queue(
                 # Offload to CPU to free VRAM; data is preserved for potential re-scoring.
                 old_extra[k] = v.cpu()
 
+    # Delete extra_outputs from batches 2+ generations behind to prevent
+    # unbounded RAM accumulation (each generation's tensors are ~4-8 GB).
+    for old_idx in list(batch_queue.keys()):
+        if old_idx < batch_index - 1 and batch_queue[old_idx].get("extra_outputs"):
+            batch_queue[old_idx]["extra_outputs"] = {}
+
```
ChuxiJ: **Bug (High): Clearing `extra_outputs` breaks LRC generation for old batches**

In `lrc_utils.py:232`, when a user navigates to any historical batch, the code accesses `batch_queue[current_batch_index]["extra_outputs"]` to retrieve tensors like `pred_latents` and `encoder_hidden_states` for LRC generation.

Clearing `extra_outputs` to `{}` for batches 2+ generations behind means that if a user navigates back to an earlier batch and clicks the LRC button, it will error because the required tensors are missing (`lrc_no_extra_outputs` / `lrc_missing_tensors`).

The original code only offloaded the previous batch's CUDA tensors to CPU (lines 49–55), freeing VRAM while preserving the data. A similar strategy is recommended:

```python
# Option A: delete only large tensors, keep non-tensor metadata
for old_idx in list(batch_queue.keys()):
    if old_idx < batch_index - 1:
        old_extra = batch_queue[old_idx].get("extra_outputs", {})
        for k in list(old_extra.keys()):
            if isinstance(old_extra[k], torch.Tensor):
                del old_extra[k]

# Option B: surface a user-visible warning or config option letting the user decide the retention policy
```

Or at minimum retain data for `batch_index - 1` and `batch_index - 2` to give users some navigation headroom.

---
_Generated by [Claude Code](https://claude.ai/code)_

acestep/core/generation/handler/init_service_memory_basic.py
```diff
@@ -105,3 +132,42 @@ def _ensure_silence_latent_on_device(self):
...
+        """Aggressively reclaim system memory after device transfers.
+
+        Combines Python GC, accelerator cache flush, and OS-level heap
+        trimming to return freed pages to the operating system.  This is
+        critical for CPU-offload workflows where PyTorch ``.to()`` creates
+        new tensor storage on each transfer and the old storage may not be
+        returned to the OS by the default C allocator.
+        """
+        gc.collect()
+        self._empty_cache()
```
ChuxiJ, line 186: **Performance (Medium): `ctypes.CDLL("libc.so.6")` reloaded on every call**

`_release_system_memory()` is called on every model load and offload, but each call executes `ctypes.CDLL("libc.so.6")` to obtain the libc handle. While ctypes has an internal cache, an explicit module-level cache is cleaner and avoids any overhead:

```python
_LIBC = None

def _get_libc():
    global _LIBC
    if _LIBC is None:
        import ctypes
        _LIBC = ctypes.CDLL("libc.so.6")
    return _LIBC
```

This lets both `_apply_malloc_mmap_threshold()` and `_release_system_memory()` reuse the same handle, eliminating repeated `import` + `CDLL` initialisation overhead.

---
_Generated by [Claude Code](https://claude.ai/code)_

acestep/core/generation/handler/init_service_memory_basic.py
```diff
@@ -1,8 +1,35 @@
...
+def _apply_malloc_mmap_threshold() -> None:
+    """Set glibc M_MMAP_THRESHOLD to 64 KB so large freed blocks go back to OS."""
+    global _MALLOPT_APPLIED
+    if _MALLOPT_APPLIED or platform.system() != "Linux":
+        return
+    try:
+        import ctypes
+        libc = ctypes.CDLL("libc.so.6")
+        # M_MMAP_THRESHOLD = -3; 64 KB threshold forces mmap for large allocs
+        libc.mallopt(-3, 65536)
```
ChuxiJ: **Note (Low): `M_MMAP_THRESHOLD=64KB` is a process-wide global setting that may affect other libraries**

Lowering the mmap threshold to 64 KB means every `malloc` call for ≥64 KB will use `mmap` instead of `sbrk`/arena allocation. This is very beneficial for releasing large PyTorch tensors, but it also affects every other library in the process (Gradio, numpy, etc.).

Frequent `mmap`/`munmap` of moderately sized blocks (64 KB–256 KB) may add extra syscall overhead. Suggestions:
1. Consider a more conservative threshold such as 128 KB or 256 KB
2. Document this global side effect so it is easy to trace if another library shows unexpected performance behaviour

---
_Generated by [Claude Code](https://claude.ai/code)_

acestep/core/generation/handler/init_service_memory_transfer.py
```diff
@@ -43,19 +44,6 @@ def _move_module_recursive(self, module, target_device, dtype=None, visited=None
...
-        for attr_name in dir(module):
-            if attr_name.startswith("_"):
-                continue
-            try:
-                attr = getattr(module, attr_name, None)
-                if isinstance(attr, torch.nn.Module) and id(attr) not in visited:
-                    self._move_module_recursive(attr, target_device, dtype, visited)
-            except (AttributeError, TypeError) as exc:
-                log = getattr(self, "logger", logger)
-                log.warning(
```
ChuxiJ, line 55: **Note (Medium): Removing the `dir()+getattr()` loop may miss non-standard submodule registrations**

The old code iterated over `dir(module)` + `getattr()` to discover submodules that were not registered via `add_module()`. The new code only traverses `module._modules`, which is correct for all standard PyTorch modules but could miss submodules stored as plain object attributes by non-standard third-party modules (i.e. not assigned via `self.xxx = nn.Module()` in `__init__`, which triggers `__setattr__` and auto-registers into `_modules`).

For ACE-Step's own model architecture this should not be a problem — standard PyTorch modules always register children automatically. However, it is worth noting this known limitation in the PR description or a code comment so that future maintainers are aware if a device-mismatch error appears after introducing a non-standard module.

---
_Generated by [Claude Code](https://claude.ai/code)_

acestep/core/generation/handler/init_service_offload_context.py
```diff
@@ -48,22 +49,37 @@ def _load_model_context(self, model_name: str):
         if model_name == "model" and hasattr(self, "silence_latent"):
             self.silence_latent = self.silence_latent.to(self.device).to(self.dtype)
 
+        # Free old CPU tensor storage after moving to GPU
+        self._release_system_memory()
+
         load_time = time.time() - start_time
```
ChuxiJ, line 52: **Minor: `_release_system_memory()` time is counted inside `load_time`**

`start_time` is recorded before `_recursive_to_device`, but `_release_system_memory()` (which includes `gc.collect()` + `malloc_trim()`) executes before `load_time` is computed. This means the reported load/offload time in logs includes GC and memory-reclaim overhead, which may mislead users into thinking transfer performance is worse than it is.

Consider moving `_release_system_memory()` to after the timer is stopped, or logging its cost separately:

```python
load_time = time.time() - start_time
self._release_system_memory()  # moved to after timing
```

---
_Generated by [Claude Code](https://claude.ai/code)_

acestep/core/generation/handler/init_service_memory_transfer.py
```diff
@@ -17,23 +21,20 @@ def _move_module_recursive(self, module, target_device, dtype=None, visited=None
...
-                    )
-                module._parameters[param_name] = moved_param
+            if param is None:
+                continue
+            if self._is_on_target_device(param, target_device):
+                if dtype is not None and param.is_floating_point() and param.dtype != dtype:
+                    param.data = param.data.to(dtype)
+                continue
+            if self._is_quantized_tensor(param):
+                module._parameters[param_name] = self._move_quantized_param(param, target_device)
```
ChuxiJ, line 32: **Bug (Low): `_move_quantized_param` still returns a new `nn.Parameter` wrapper**

For regular parameters, this PR switched to `param.data = new_data` to avoid creating a new `nn.Parameter` wrapper (reducing memory allocations). However, the quantized-parameter path still calls `self._move_quantized_param()`, which returns a `torch.nn.Parameter(...)` new object that is then assigned to `module._parameters[param_name]`.

This does not affect correctness, but it is inconsistent with the "use in-place `param.data` replacement" optimisation described in the PR. If the underlying storage of quantized parameters can also be modified in-place via `_apply_fn_to_data`, it may be worth unifying the two paths.

---
_Generated by [Claude Code](https://claude.ai/code)_

ChuxiJ:
## Code Review Fixes Submitted (484723b)

Fixes for all issues identified in the review have been pushed to the `claude/gallant-heisenberg` branch:

### 🔴 Bug Fix

**1. `batch_queue` cleanup no longer breaks LRC generation** (`batch_queue.py`)
- **Problem**: The original implementation cleared the entire `extra_outputs` dict to `{}` for old batches, causing errors when users navigated back to an older batch and clicked the LRC button (tensors like `pred_latents` were missing)
- **Fix**: Now only deletes entries whose value is a `torch.Tensor` (the large objects), preserving non-tensor metadata (e.g. `lm_metadata`, `time_costs`) so both memory is freed and batch navigation continues to work

### 🟡 Performance & Quality Improvements

**2. Cache the libc handle** (`init_service_memory_basic.py`)
- **Problem**: `_release_system_memory()` called `ctypes.CDLL("libc.so.6")` on every invocation
- **Fix**: Added a `_get_libc()` helper that caches the handle globally, shared by both `_apply_malloc_mmap_threshold()` and `_release_system_memory()`

**3. Raise M_MMAP_THRESHOLD to 128 KB** (`init_service_memory_basic.py`)
- **Problem**: The 64 KB threshold was too aggressive and could cause frequent mmap/munmap syscalls for medium-sized allocations
- **Fix**: Changed threshold to 128 KB and documented in the docstring that this is a process-wide global setting

**4. Fix load/offload timing** (`init_service_offload_context.py`)
- **Problem**: `_release_system_memory()` (including `gc.collect()` + `malloc_trim()`) was counted inside `load_time`/`offload_time`, inflating the reported transfer latency in logs
- **Fix**: Moved `_release_system_memory()` calls to after the timer stops

**5. Add `_modules` traversal limitation note** (`init_service_memory_transfer.py`)
- Documents that `_move_module_recursive` only visits submodules registered in `_modules`; attributes stored as plain object properties (not via `add_module`) will not be moved

### ✅ Test Verification

All 59 `init_service_test.py` tests pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

ChuxiJ:
## Added `save_memory_mode` Global Setting (507ae50)

Added a global toggle `save_memory_mode` that, when enabled, completely avoids storing intermediate tensors in `extra_outputs` (approximately 4–8 GB per generation), eliminating the memory issue at the source.

### Usage

```bash
# Enable via environment variable (disabled by default)
ACESTEP_SAVE_MEMORY=1 python acestep
```

### Behavior Changes

| Feature | Default mode | save_memory_mode |
|---------|-------------|-----------------|
| Tensors in `extra_outputs` | Retained (`pred_latents`, `encoder_hidden_states`, etc.) | Not stored; only metadata retained (`time_costs`, `spans`, `seed_value`) |
| Auto Score checkbox | Interactive | Greyed out / non-interactive |
| Auto LRC checkbox | Interactive | Greyed out / non-interactive |
| Manual LRC button | Works normally | Shows i18n info message |
| `auto_score`/`auto_lrc` during generation | Follows user selection | Forcibly skipped |

### Changed Files

| File | Change |
|------|--------|
| `gpu_config.py` | Added `save_memory_mode: bool = False` field to `GPUConfig` + `ACESTEP_SAVE_MEMORY` env var support |
| `generate_music_payload.py` | Skips all `.detach().cpu()` tensor copies when save_memory_mode is on |
| `generation_progress.py` | Forces `auto_score = False`, `auto_lrc = False` |
| `generation_tab_generate_controls.py` | UI checkboxes set to `interactive=not save_memory` |
| `lrc_utils.py` | Manual LRC button early-returns with i18n info message |
| `i18n/{en,zh,ja,he}.json` | Added `lrc_save_memory_disabled` message key |

### Tests

665 tests pass (4 failures + 15 errors are all pre-existing issues, unrelated to this change).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1091, opened 2026-04-12
fix: add ACESTEP_SKIP_VRAM_PREFLIGHT env var + gc before preflight check

FlexOr2:
### Summary
- Run `gc.collect()` + `torch.cuda.empty_cache()` before the VRAM pre-flight check so it measures actual free VRAM, not memory held by PyTorch's caching allocator
- Add `ACESTEP_SKIP_VRAM_PREFLIGHT=1` env var to bypass the check entirely for tight-VRAM setups

### Scope
- **File changed:** `acestep/core/generation/handler/generate_music.py`
- **Out of scope:** `_vram_preflight_check()` method itself is unchanged

### Risk and Compatibility
- **Target platform:** CUDA (shared-GPU setups with 24 GB cards where desktop compositor shares VRAM)
- **Non-target platforms unchanged:** CPU/MPS/XPU paths already return None before reaching the preflight check
- **Default behavior unchanged:** check runs as before unless the env var is explicitly set

### Why
On shared-GPU setups (e.g., 24 GB RTX 3090 with desktop compositor using 2-3 GB), the pre-flight check reports e.g. "1.3 GB free, needs 1.4 GB" and blocks generation, even though PyTorch's caching allocator can handle it via `expandable_segments:True`. The `gc.collect()` + `torch.cuda.empty_cache()` before the check reclaims fragmented allocator cache so the measurement reflects actual free memory. The env var provides an escape hatch for setups where even the corrected check is too conservative.

### Regression Checks
- Verified on RTX 3090 (24 GB) with desktop compositor (2.5 GB shared)
- Default path (no env var): check still runs, now with more accurate free VRAM measurement
- `ACESTEP_SKIP_VRAM_PREFLIGHT=1`: check skipped, debug log emitted
- Existing `_vram_preflight_check` unit tests pass (method unchanged)

### Reviewer Notes
- Follows existing env var convention: `ACESTEP_` prefix, boolean parsing via `.lower() in ("1", "true", "yes")`
- Similar pattern to `ACESTEP_VAE_ON_CPU` in `generate_music_decode.py`

## Summary by CodeRabbit

* **New Features**
  * Expanded music generation with DCW configuration, cached repaint latents, retake seed and variance controls, and flow-edit parameters.
  * Added support for flow editing with source audio duration handling across supported generation modes.
  * Added an environment-driven option to bypass GPU VRAM preflight checks when needed.
* **Bug Fixes**
  * Improved default handling by resolving unspecified DCW settings from the loaded model.
  * Added warnings for unsupported flow-edit task types.

### DIFF
--- a/acestep/core/generation/handler/generate_music.py
+++ b/acestep/core/generation/handler/generate_music.py
@@ -5,6 +5,7 @@ This module provides the public ``generate_music`` entry point extracted from
 """
 
 import gc
+import os
 import traceback
 from typing import Any, Dict, List, Optional, Union
 
@@ -386,13 +387,24 @@ class GenerateMusicMixin:
                 repainting_end=repainting_end,
                 chunk_mask_mode=chunk_mask_mode,
             )
-            vram_error = self._vram_preflight_check(
-                actual_batch_size=actual_batch_size,
-                audio_duration=audio_duration,
-                guidance_scale=guidance_scale,
-            )
-            if vram_error is not None:
-                return vram_error
+            gc.collect()
+            torch.cuda.empty_cache()
+            skip_preflight = os.environ.get(
+                "ACESTEP_SKIP_VRAM_PREFLIGHT", "",
+            ).lower() in ("1", "true", "yes")
+            if skip_preflight:
+                logger.debug(
+                    "[generate_music] VRAM pre-flight skipped "
+                    "(ACESTEP_SKIP_VRAM_PREFLIGHT=true)"
+                )
+            else:
+                vram_error = self._vram_preflight_check(
+                    actual_batch_size=actual_batch_size,
+                    audio_duration=audio_duration,
+                    guidance_scale=guidance_scale,
+                )
+                if vram_error is not None:
+                    return vram_error
 
             injection_ratio, resolved_cf_frames, resolved_wav_cf = (
                 _resolve_repaint_config(repaint_mode, repaint_strength)
--- a/acestep/core/generation/handler/generate_music.py
+++ b/acestep/core/generation/handler/generate_music.py
@@ -387,25 +387,26 @@ class GenerateMusicMixin:
                 repainting_end=repainting_end,
                 chunk_mask_mode=chunk_mask_mode,
             )
-            gc.collect()
-            torch.cuda.empty_cache()
-            skip_preflight = os.environ.get(
-                "ACESTEP_SKIP_VRAM_PREFLIGHT", "",
-            ).lower() in ("1", "true", "yes")
-            if skip_preflight:
-                logger.warning(
-                    "[generate_music] VRAM pre-flight check skipped via "
-                    "ACESTEP_SKIP_VRAM_PREFLIGHT=1. If generation OOMs, "
-                    "unset this variable to re-enable the safety check."
-                )
-            else:
-                vram_error = self._vram_preflight_check(
-                    actual_batch_size=actual_batch_size,
-                    audio_duration=audio_duration,
-                    guidance_scale=guidance_scale,
-                )
-                if vram_error is not None:
-                    return vram_error
+            if torch.cuda.is_available():
+                gc.collect()
+                torch.cuda.empty_cache()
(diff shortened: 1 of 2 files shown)


ChuxiJ (commented):
Thanks @FlexOr2 — nice, surgical patch. I traced through the surrounding code (`generate_music.py:105-163` for `_vram_preflight_check` and the call site at `:329`) and it's correct in all the ways I checked:

- `gc.collect()` **paired with** `torch.cuda.empty_cache()` is the right idiom here, not cargo-cult. Without the GC pass, `empty_cache()` can't return blocks that still have live Python references; without `empty_cache()`, `gc.collect()` alone leaves PyTorch's caching allocator holding the memory. This matches existing patterns elsewhere in the repo (e.g. `llm_inference.py:107,136,584`, `init_service_loader.py:134`, `reinitialize_route.py:99`).
- Safe on non-CUDA platforms. I verified `torch.cuda.empty_cache()` is a silent no-op on Mac (no CUDA available), and `_vram_preflight_check` itself early-returns at line 128 when `torch.cuda.is_available()` is False, so CPU/MPS/XPU see no behavioral change beyond a cheap extra GC pass.
- Env-var naming (`ACESTEP_SKIP_VRAM_PREFLIGHT`) and boolean parsing (`.lower() in ("1","true","yes")`) are consistent with the project's existing `ACESTEP_*` conventions.
- Required imports (`gc`, `torch`, `logger`) were already present; the diff only adds `import os` which is correct.

One change I'd like before this lands:

### Please raise the skip notice from `debug` to `warning`

```python
if skip_preflight:
    logger.debug(
        "[generate_music] VRAM pre-flight skipped "
        "(ACESTEP_SKIP_VRAM_PREFLIGHT=true)"
    )
```

`logger.debug` is filtered out at the default log level, so a user who sets `ACESTEP_SKIP_VRAM_PREFLIGHT=1` gets no feedback that the safety net is off. The bigger problem is that if a later generation OOMs, nothing in the logs explains *why* preflight didn't catch it — support/debugging becomes harder because there's no breadcrumb.

Suggested change:

```python
if skip_preflight:
    logger.warning(
        "[generate_music] VRAM pre-flight check skipped via "
        "ACESTEP_SKIP_VRAM_PREFLIGHT=1. If generation OOMs, "
        "unset this variable to re-enable the safety check."
    )
```

This way:
- Users who intentionally set the env var see an explicit reminder each run that they're running without the safety net (intentional, healthy friction).
- OOM post-mortems have a clear breadcrumb.
- Still non-blocking — we're not spamming errors, just one warning line per generation.

### Optional nit (not blocking)

You could move the `gc.collect()` / `empty_cache()` pair into the `else:` branch so we don't pay the GC cost when the check is skipped anyway:

```python
skip_preflight = os.environ.get("ACESTEP_SKIP_VRAM_PREFLIGHT", "").lower() in ("1", "true", "yes")
if skip_preflight:
    logger.warning(...)
else:
    gc.collect()
    torch.cuda.empty_cache()
    vram_error = self._vram_preflight_check(...)
    if vram_error is not None:
        return vram_error
```

Not a blocker — the GC overhead is negligible in practice (music generation runs on the order of seconds), and keeping it outside has the small advantage of also clearing memory before the actual generation path even if the preflight is skipped. Totally your call.

Happy to approve once the `logger.debug` → `logger.warning` change is in. Nice diagnosis on the root cause in the PR description — the `expandable_segments:True` note in particular helped me understand why the measurement-vs-reality gap is platform-specific.

FlexOr2:
Rebased onto upstream main at 14c0211. Range-diff confirms the three existing fix commits are semantically unchanged; added one focused test commit for default-enabled versus explicit opt-out behavior. Targeted result: 12 passed; Ruff clean. New fork head: 2c0abcb.

ChuxiJ (approved):
Re-reviewed the latest head after the warning/CUDA-gating follow-ups. The change is now scoped to CUDA paths, the skip notice is visible, and focused verification passed locally:  (12 tests).

PR #1092, opened 2026-04-12
feat(api): expose sampler_mode and 4 DiT params on /release_task HTTP API

FlexOr2:
### Summary
- Expose 5 existing `GenerationParams` fields on the `/release_task` HTTP endpoint
- These params were only accessible via Gradio's direct Python call path, not through the HTTP API used by external clients

### Parameters added
| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `sampler_mode` | str | `"euler"` | `"euler"` or `"heun"` sampler |
| `velocity_norm_threshold` | float | `0.0` | Velocity norm clamping (0=off) |
| `velocity_ema_factor` | float | `0.0` | Velocity EMA smoothing (0=off) |
| `latent_shift` | float | `0.0` | Additive shift on latents before VAE decode |
| `latent_rescale` | float | `1.0` | Multiplicative rescale on latents before VAE decode |

### Scope
- `acestep/api/http/release_task_models.py` — add fields to `GenerateMusicRequest`
- `acestep/api/http/release_task_param_parser.py` — add camelCase aliases
- `acestep/api/http/release_task_request_builder.py` — wire through builder
- `acestep/api/job_generation_setup.py` — pass to `GenerationParams`
- **Out of scope:** Gradio UI, training, CLI, GenerationParams itself (unchanged)

### Risk and Compatibility
- **Fully backward compatible**: all 5 fields have defaults matching existing `GenerationParams` defaults. Omitting them preserves current behavior.
- **Non-target paths unchanged**: Gradio UI continues to call `generate_music()` directly; this change only affects the HTTP API surface.
- `getattr()` with defaults in `job_generation_setup.py` ensures older request objects (without the new fields) still work.

### Why
External clients using the HTTP API (e.g., custom UIs, automation scripts) cannot currently control the sampler mode or latent post-processing params. The Gradio UI bypasses the HTTP API entirely (calling Python directly), so these params work there but not via `/release_task`. This creates a capability gap between Gradio and HTTP API users.

### Regression Checks
- Default behavior unchanged when params are omitted
- `PARAM_ALIASES` entries follow existing camelCase convention
- Existing `release_task_param_parser_test.py` tests unaffected (no existing test touches these new fields)

## Summary by CodeRabbit

* **New Features**
  * Added advanced sampling and latent-space controls for music generation, including Euler/Heun sampler selection, velocity settings, latent shifting, and latent rescaling.
  * New parameters support both snake_case and camelCase request formats with backward-compatible defaults.

* **Tests**
  * Added coverage for parameter aliases, request propagation, and default values.


FlexOr2:
Rebased onto upstream main at 14c0211. Range-diff confirms the two existing API commits are semantically unchanged. The targeted run exposed a stale upstream parser test stub; the follow-up commit aligns that stub and removes the now-unused import. Targeted result: 24 passed plus 4 subtests; Ruff clean. New fork head: d267a52.

ChuxiJ (changes requested):
The focused API tests pass locally (24 tests), and the parameter plumbing is generally in the right place. I would like one API-boundary fix before this lands.

`sampler_mode` is documented as only `"euler"` or `"heun"`, but the request model currently accepts any string. That gives inconsistent behavior downstream: the MLX path validates and raises, while the PyTorch/Base path effectively treats non-`"heun"` values as Euler. Since this PR exposes the field on the public HTTP API, please validate it at the request boundary, e.g. `Literal["euler", "heun"]`, and add an invalid-value regression test.

While touching the tests, please also keep the existing omitted-field/default coverage from the latest CodeRabbit note. After #1282 and #1091 landed today, a quick rebase onto current `main` would also make this easier to merge cleanly.

Issue #1081, opened 2026-04-09
Auto-score on MacOS with MLX with XL-Turbo and and 5HZ-4B Models Loaded Drains Memory


lbryant2004-create:
Hello, I wanted to point out what could possibly be a bug on MacOS:
**Did you check DeepWiki?**
- [ ] No, this is a new bug not covered in the docs

To avoid vroom errors, use export PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 and export PYTORCH_ENABLE_MPS_FALLBACK=1
(When I try to use ACE-Step with neither of these set on my 32GB RAM system I get vroom errors.)
Run ACE-Step version 82252c2).
Load the acestep-v15-xl-turbo and the acestep-5Hz-lm-4B models into memory. If you want to keep this debug simple, choose the Simple Radio Button and leave the caption for a random result and press the Create Sample button. Or, from Custom Mmode type in a caption and lyrics if you wish, but this bug also shows up if you're creating an instrumental. To see the bug, make sure the Autoscore option is checked and hit Generate Music after setting optional parameters. What happens is that once the MP3 or audio file is generated, you'll see the Terminal reloading the models, (using even more memory according to Activity Monitor) then it does the scoring. If Autoscore isn't checked, the generation is complete and no memory is lost. If it is checked, if you don't crash rightaway, try another generation with autoscore and eventually you'll run out of all memory causing the system to crash.

**Expected behavior**
The models shouldn't need to be reloaded, the scoring should be able to take place just fine.

**Desktop (please complete the following information):**
OS: Tahoe 26.3.1 (a) MAC Mini 32GB RAM ARM m4 chip 512GB SSD
Google Chrome- Version 146.0.7680.178 (Official Build) (arm64)

ChuxiJ:
Thanks for the detailed report — the symptoms you describe match what I found after walking through the Autoscore code paths. Here's my analysis of the root cause and the fix I plan to propose.

## Root cause

On the **MLX backend**, enabling **Autoscore** triggers two expensive memory paths on every generation, and at least one of them leaks permanently on unified-memory Macs:

### 1. A full second copy of the 5Hz-lm-4B is loaded as a HuggingFace PyTorch model

[...]

PR #1097, opened 2026-04-13
fix(scoring): stop Autoscore unified-memory leak on MLX (issue #1081)

ChuxiJ:
Fixes the catastrophic unified-memory growth described in #1081, where enabling **Autoscore** on macOS with the MLX backend (`acestep-v15-xl-turbo` + `acestep-5Hz-lm-4B` loaded, 32 GB Mac) would drain RAM and crash the system after only a few generations.

## Root cause

On the MLX backend, Autoscore triggers three compounding memory paths:

1. **A second full copy of the LM is loaded as a HuggingFace PyTorch model.** The MLX weights cannot be used for PyTorch teacher-forcing scoring, so `get_hf_model_for_scoring()` (`acestep/llm_inference.py:4163-4194`) calls `AutoModelForCausalLM.from_pretrained(...)` and caches the result on `self._hf_model_for_scoring`. For the 5Hz-lm-4B model that is roughly **~8 GB of CPU-resident memory that never gets freed**.

2. **That cached model was migrated CPU ↔ MPS on every forward pass.** `_load_scoring_model_context` in `acestep/core/scoring/lm_score.py` moved the model to the accelerator and back inside every call. One Autoscore pass performs dozens of forward passes per sample (per-metadata recall + caption PMI + lyrics PMI, each with a conditional and unconditional prompt), so the MPS allocator accumulated fragmented multi-GB allocations every time.

3. **`PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0` removes the last guardrail.** That env var is required to even start the app with both models loaded on a 32 GB Mac, but it disables the MPS allocator's high-watermark check, letting the fragmentation above push the process past unified-memory limits.

Disabling Autoscore makes the leak disappear entirely, because `generation_progress.py:280` gates the whole scoring block behind `if auto_score:` — that matches the reporter's observation.

## Fix

Two localised edits in `acestep/core/scoring/lm_score.py`:

### 1. `_load_scoring_model_context` is now reentrant
Nested entries are tracked via a depth counter (`llm_handler._scoring_ctx_depth`) and become no-ops. Only the outermost entry performs the CPU → accelerator migration; only the outermost exit performs the offload. This mirrors the device-based reentrancy guard that already exists for the `pt` backend in `llm_inference.py:4075-4082`.

### 2. `calculate_pmi_score_per_condition` batches all forward passes under one outer context
The whole scoring body is now wrapped in a single `with _load_scoring_model_context(llm_handler):`. All the inner `_get_logits_and_target_for_scoring` calls still enter the context, but the reentrancy guard turns them into no-ops. Result: **exactly one CPU ↔ MPS migration per Autoscore pass, instead of one per forward pass.**

### 3. MLX-specific: drop the cached HF copy on outermost exit
When the backend is MLX and `offload_to_cpu` is enabled, the outermost exit additionally:
- sets `llm_handler._hf_model_for_scoring = None`
- runs `gc.collect()`
- empties the MPS cache

This returns the ~8 GB duplicate PyTorch copy to the OS between scoring passes. The next scoring call re-materialises the model via `AutoModelForCausalLM.from_pretrained`; HF hub weight caching makes this a fast rematerialisation from disk, not a redownload. A 32 GB Mac pays a one-time few-seconds reload cost on each Autoscore run in exchange for not crashing.

Other backends are unaffected:
- `pt`: still delegates to `llm_handler._load_model_context()` which already has its own reentrancy guard.
- `vllm`: same hoisted outer context, but the HF scoring model is kept cached (the CUDA allocator handles this cleanly and VRAM is not unified with system RAM).

## Files changed
- `acestep/core/scoring/lm_score.py` — +110 / -40

## Test plan
- [ ] macOS unit memory test: on a 32 GB Apple Silicon Mac with MLX backend, load `acestep-v15-xl-turbo` + `acestep-5Hz-lm-4B`, enable Autoscore, run ≥5 consecutive generations. Expected: peak RSS plateaus between generations instead of growing monotonically; no crash.
- [ ] Check Terminal logs show `[scoring] Released cached HF scoring model on MLX backend` after each Autoscore pass.
- [ ] Verify PMI scores are numerically identical before/after (the reentrant guard should be behaviour-preserving).
- [ ] Smoke-test PyTorch backend: run Autoscore on CUDA and confirm no behavioural change (delegates to `_load_model_context()` as before).
- [ ] Smoke-test vllm backend (if available): confirm Autoscore still works with the hoisted outer context.

## Follow-ups (not in this PR)
- Per-sample DiT migration in `get_lyric_score` (`acestep/core/generation/handler/lyric_score.py:93`) is still once-per-sample. A future PR could hoist a single `_load_model_context("model")` around the whole Autoscore loop in `generation_progress.py:280`.
- UI-side warning when MLX + ≤32 GB unified memory + both models loaded + Autoscore enabled (extension of the `acestep/gpu_config.py:261` heuristic).

Refs: #1081

https://claude.ai/code/session_01Wr7kjcUxgdkf5oUATzHY9U

## Summary by CodeRabbit

* **Performance**
  * Reduced redundant model transfers during scoring: model migration to/from accelerator now happens once per autoscore pass, lowering runtime and accelerator memory usage and avoiding races under concurrency; cached accelerator memory is explicitly cleared when offloaded.
* **Tests**
  * Added regression tests validating scoring context lifecycle, nesting, cross-thread behavior, and correct load/offload semantics.

### DIFF
--- a/acestep/core/scoring/lm_score.py
+++ b/acestep/core/scoring/lm_score.py
@@ -9,6 +9,7 @@ import contextlib
 import gc
 import math
 import re
+import threading
 
 import torch
 import torch.nn.functional as F
@@ -68,19 +69,64 @@ def pmi_to_normalized_score(pmi: float, scale: float = 0.1) -> float:
     return 1.0 / (1.0 + math.exp(-pmi / scale))
 
 
-def _empty_accelerator_cache() -> None:
-    """Release cached GPU/MPS memory on the active accelerator."""
-    if torch.cuda.is_available():
-        torch.cuda.empty_cache()
-    elif hasattr(torch, "xpu") and torch.xpu.is_available():
-        torch.xpu.empty_cache()
-    elif (
-        hasattr(torch.backends, "mps")
-        and torch.backends.mps.is_available()
-        and hasattr(torch, "mps")
-        and hasattr(torch.mps, "empty_cache")
-    ):
-        torch.mps.empty_cache()
+def _empty_accelerator_cache(backend: str) -> None:
+    """Release cached accelerator memory scoped to the scoring backend.
+
+    Keeps cache eviction confined to the runtime actually used by the
+    scoring pass so non-target runtimes (e.g. CUDA when running MLX) are
+    not forced through an allocator reset.  ``mlx`` -> MPS, ``vllm`` ->
+    CUDA / XPU, ``pt`` -> delegated to ``_load_model_context`` and never
+    reaches this helper.
+    """
+    if backend == "mlx":
+        if (
+            hasattr(torch.backends, "mps")
+            and torch.backends.mps.is_available()
+            and hasattr(torch, "mps")
+            and hasattr(torch.mps, "empty_cache")
+        ):
+            torch.mps.empty_cache()
+        return
+
+    if backend == "vllm":
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
+        elif hasattr(torch, "xpu") and torch.xpu.is_available():
+            torch.xpu.empty_cache()
+
+
+# Per-thread reentrancy depth state for _load_scoring_model_context.  Keeping
+# this on ``threading.local`` (rather than on the handler) prevents concurrent
+# Autoscore calls on the same handler from seeing each other's depth counter
+# and clearing the cached scoring model out from under a peer thread.  The
+# scoring forward path is strictly synchronous (``torch.no_grad`` + a plain
+# ``model(...)`` call) so thread-local storage is sufficient; there are no
+# async suspension points where a contextvar would be strictly required.
+_scoring_tls = threading.local()
+
+# Guards lazy creation of the per-handler RLock so two threads racing on the
+# first scoring call don't each install a different lock object.
+_scoring_lock_creation_lock = threading.Lock()
+
+
+def _get_scoring_handler_lock(llm_handler) -> threading.RLock:
+    """Return (and lazily create) a per-handler reentrant lock."""
+    lock = getattr(llm_handler, "_scoring_lock", None)
+    if lock is None:
+        with _scoring_lock_creation_lock:
+            lock = getattr(llm_handler, "_scoring_lock", None)
+            if lock is None:
+                lock = threading.RLock()
+                llm_handler._scoring_lock = lock
+    return lock
+
+
+def _scoring_depth_map() -> Dict[int, int]:
+    depths = getattr(_scoring_tls, "depths", None)
+    if depths is None:
+        depths = {}
+        _scoring_tls.depths = depths
+    return depths
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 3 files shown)


ChuxiJ:
## Review: verified on macOS, one issue found and fixed

I read the full diff against `main`, traced the call graph, and ran empirical tests on a Mac. Here's what I found:

### ✅ Cross-thread stale-model race — **NOT a real bug**

I wrote a test with two threads racing into `_load_scoring_model_context` on a shared handler. Event-log output confirmed:

```
A: entering → A: inside_ctx → B: entering (BLOCKED) → A: exiting → B: inside_ctx
```

Thread B is blocked at the `RLock` acquisition until Thread A fully exits. The `with lock:` at L202 wraps the entire `try: yield` block, so the lock is held across the yield. This makes the race condition described in a prior code-review pass unreachable — the RLock serializes the shared `_hf_model_for_scoring` load/offload transition correctly.

The existing `test_outermost_detection_is_thread_local` test also passed on macOS with both threads correctly identified as outermost (get_calls == 2).

### ❌ Exception-path cleanup — **was a real bug, now fixed**

I injected a `RuntimeError` into `model.to("cpu")` and observed that `_hf_model_for_scoring` was **not** cleared afterwards — the MLX release path at L234+ was skipped because the exception propagated out of the `finally` block before reaching it.

Reproduction:

```python
class FailingModel:
    def to(self, device):
        if device == "cpu":
            raise RuntimeError("simulated offload failure")
        return self

handler = Handler("mlx", offload=True)
handler._hf_model_for_scoring = FailingModel()
with _load_scoring_model_context(handler):
    pass
# → _hf_model_for_scoring is still cached (should be None)
```

**Fix pushed in `1f03d25`**: wrapped the CPU offload in a nested `try/finally` so the MLX-specific release (`_hf_model_for_scoring = None`, `gc.collect()`, `torch.mps.empty_cache()`) runs regardless of whether the offload succeeds. Added `test_mlx_cleanup_runs_when_offload_raises` as a regression test — all 8 tests pass on macOS.

### Test results

```
8 passed in 1.11s
```

All original 7 tests + the new exception-path test pass on macOS (Python 3.11.4, Apple Silicon).

### Other verification notes

- `_empty_accelerator_cache(backend)` correctly scopes cache eviction to the active backend (mlx → MPS, vllm → CUDA/XPU). No cross-backend interference.
- The only external entry point is `scoring.py:97` → `calculate_pmi_score_per_condition`, which always wraps all inner calls in one outer `_load_scoring_model_context`. No call site bypasses the outer wrapper.
- RLock serializing vllm too is intentional and acceptable per the PR description.
- Thread-local depth tracking via `_scoring_tls` keyed by `id(llm_handler)` is correct for this use case (synchronous `torch.no_grad` forward pass, no async suspension points).

ChuxiJ:
PR #1170 now includes a disk-backed Gradio feature cache for the Auto Score / Auto LRC tensor-retention path: per-sample alignment tensors are persisted under the generated source session (or feature_sessions) and batch history keeps cache paths instead of large tensors. I referenced this PR from #1170 rather than using a closing keyword because #1097 also contains a separate MLX scoring-model lifecycle optimization.

PR #1123, opened 2026-04-21
feat(api): expose APG eta and momentum on /release_task HTTP API

FlexOr2:
## Summary

Surfaces the two remaining APG (Adaptive Projected Guidance) knobs — `eta` and `momentum` — on the `/release_task` HTTP API, following the pattern of #1092.

## Stacking note

This branch is based on the tip of #1092, not on `main`, because both PRs touch `release_task_models.py`, `release_task_param_parser.py`, `release_task_request_builder.py`, and `job_generation_setup.py`. Filing as a Draft — I'll rebase onto `main` and mark ready-for-review once #1092 merges. If you'd prefer I close this and refile after #1092 lands, happy to.

## Context

Public community workflows (e.g. the [Civitai ACE-Step 1.5 XL workflow](https://civitai.com/models/2375403)) tune APG using `eta` (parallel-component scaling, currently hardcoded to `0.0` at `acestep/models/common/apg_guidance.py:38`) and `momentum` (the `MomentumBuffer` decay factor, currently hardcoded to `-0.75` at `acestep/models/common/apg_guidance.py:7`). Both are frozen at the `MomentumBuffer()` / `apg_forward()` call sites inside each DiT model, so HTTP API clients can't tune them.

#1092 established the pattern for exposing internal DiT params on `/release_task`. This PR reuses that pattern for `eta` and `momentum` so HTTP clients reach parity with ComfyUI users for APG tuning.

## Changes

- `GenerationParams` gains `eta: float = 0.0` and `momentum: float = -0.75` (both defaults match the previous hardcoded values — see Backward compatibility below).
- `generate_audio` on the four DiT model variants that use APG — `base`, `sft`, `xl_base`, `xl_sft` — accepts `eta` and `momentum` as keyword-only params and forwards them to `MomentumBuffer(momentum=...)` and `apg_forward(..., eta=...)`.
- Turbo variants (`turbo`, `xl_turbo`) are distilled without CFG → no APG path → not affected.
- The kwargs thread through `generate_music` → `_run_generate_music_service_with_progress` → `service_generate` → `_build_service_generate_kwargs`.
- HTTP surface: `GenerateMusicRequest`, `PARAM_ALIASES`, `build_generate_music_request`, and `build_generation_setup` all accept and forward the new fields.
- Unit tests cover alias resolution, Pydantic defaults, request-builder forwarding, `GenerationSetup` wiring, and a model-level invariant that omitting the new kwargs yields the same output as passing the historical defaults explicitly.

[...]

PR #1129, opened 2026-04-22
refactor: pluggable guidance variant registry

### LINKED ISSUE #1124: RFC: pluggable guidance function registry for sampler customization
# RFC: pluggable guidance function registry for sampler customization

## TL;DR

- **Problem.** `apg_guidance.py` already holds 5 guidance variants (APG, CFG, and 3 ADG flavors), but selection is hardcoded in every model and every new APG knob needs its own PR — cf. #1092 plus a follow-up I've prepared for `eta` + `momentum`.
- **Proposal.** Adopt a `GuidanceFn` registry — same DI pattern as PyTorch hooks, Diffusers `callback_on_step_end`, HF `LogitsProcessor`, ComfyUI nodes. ~200–300 lines, PyTorch-only, zero behavior change by default.
- **Ask.** Interest check before I invest in a Draft PR. "No" is a perfectly fine answer — I'll continue with one-knob-per-PR contributions.

## Context

[`acestep/models/common/apg_guidance.py`](acestep/models/common/apg_guidance.py) already ships three core guidance algorithms and two ADG wrapper variants — all in one file:

| Function | Line | Role |
|---|---|---|
| `apg_forward` | 33–56 | Adaptive Projected Guidance |
| `cfg_forward` | 59–60 | Classic classifier-free guidance |
| `adg_forward` | 107–188 | Angle-based Diffusion Guidance |
| `adg_w_norm_forward` | 191–212 | ADG + norm preservation |
| `adg_wo_clip_forward` | 215–228 | ADG without angle clip |

Selection between them is **hardcoded** inside each model's `generate_audio()`. For example [`modeling_acestep_v15_base.py:2025-2039`](acestep/models/base/modeling_acestep_v15_base.py) calls `apg_forward` in one branch and `adg_forward` in another, both with fixed parameter choices. `MomentumBuffer()` is instantiated inline at `:1943` with the default `momentum=-0.75` baked in.

So the architecture is **already halfway to pluggable**: multiple variants exist and get selected, but selection logic is frozen and no public entry point lets external users add their own variant (e.g. APG-with-cosine-scheduled-eta, APG-OR, or future paper proposals) without forking and editing each of the four APG-using model files.

The HTTP API inherits this rigidity: every new APG knob needs a dedicated PR — cf. #1092 exposing 5 sampler/DiT params, and a follow-up PR I've prepared for `eta` + `momentum` because they're not plumbed through to the HTTP layer either. Each such PR is small but the pattern will repeat indefinitely as new guidance variants are published.

## Precedent in this repo

Maintainers have previously done consolidation/DRY work in this area: **#1000** ("chore: deduplicate model config and guidance files") and **#998** ("chore: deduplicate exact-copy files across model variants"). A registry would be the natural next step on that same consolidation trajectory.

## Proposed pattern

Adopt the standard dependency-injection convention used across the Python ML ecosystem:

- **PyTorch:** [`register_forward_hook`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_hook)
- **Diffusers:** `callback_on_step_end` accepts user callables
- **Transformers:** `StoppingCriteria` / `LogitsProcessor` registries
- **ComfyUI:** every node is this pattern, just exposed as a graph

A sketch (signature is a draft for discussion — would refine based on your feedback):

```python
# acestep/models/common/guidance_registry.py (new)
from typing import Protocol, Callable
import torch

class GuidanceFn(Protocol):
    def __call__(
        self,
        pred_cond: torch.Tensor,
        pred_uncond: torch.Tensor,
        guidance_scale: float,
        state: dict,              # owned by the sampler loop, cleared each generate_audio() call
        **params,                 # per-variant params (eta, norm_threshold, angle_clip, etc.)
    ) -> torch.Tensor: ...

_REGISTRY: dict[str, GuidanceFn] = {
    "apg_classic": guidance_apg_classic,   # wraps apg_forward
    "cfg":         guidance_cfg,           # wraps cfg_forward
    "adg":         guidance_adg,           # wraps adg_forward
    "adg_w_norm":  guidance_adg_w_norm,
    "adg_wo_clip": guidance_adg_wo_clip,
}

def register_guidance(name: str) -> Callable[[GuidanceFn], GuidanceFn]:
    """Decorator for third-party variants."""
    ...
```

HTTP surface (arbitrary callables can't serialize, so use registered-name + param dict):

```json
{
  "guidance_variant": "apg_classic",
  "guidance_params": {"eta": 1.05, "norm_threshold": 1.3, "momentum": -0.25}
}
```

`generate_audio()` takes one extra optional keyword: `guidance_fn: Callable | None = None`. Default preserved → current behavior byte-for-byte unchanged.

## Scope estimate

- **PyTorch-only initially.** MLX `_mlx_apg_forward` has a different state-dict shape and no `eta` parameter; unifying is separate work.
- Registry file: ~80–120 new lines.
- Dispatch sites in the 4 APG-using model variants: ~5 lines each → ~20 lines modified.
- Tests for the registry + each wrapper: ~100–150 lines.
- **Total: ~200–300 lines, 6–8 files.** Aligned with "Minimize Blast Radius" if kept PyTorch-only and without touching any existing test.

## Honest tradeoffs

I owe an accurate picture of the cost, not just the upside:

**Benefits for upstream:**
- Less per-variant maintenance. Community-submitted variants land as small, contained PRs adding one registry entry instead of editing four model files each time.
- Closes the HTTP-vs-ComfyUI knob-parity gap. HTTP clients stop needing `/release_task` patches for every new guidance recipe.
- Research-friendly: your own researchers can prototype new guidance variants without touching the sampler core.

**Real costs:**
- **API commitment.** Once `GuidanceFn` is public, refactoring the internal signature becomes a semver-breaking change.
- **Debugging blast radius.** A user's custom guidance has a bug → looks like inference quality issue → your support triage gets harder. Mitigatable with a "user-registered variant" log stamp.
- **Documentation burden.** Per-variant docstrings + recommended param ranges + edge-case warnings. Not huge, but real.
- **Testing surface expansion.** `apg_classic`, `cfg`, `adg`, and the two ADG wrappers all become public contracts requiring regression coverage.
- **Two subtle corners that need explicit decisions before code lands:**
  1. **CFG=1 edge case.** When `guidance_scale == 1.0`, most variants are no-ops but momentum state may still initialize and accumulate. Harmless but wasteful; contract should specify.
  2. **State-dict lifetime.** Currently `MomentumBuffer` is scoped to one `generate_audio()` call. A plugin pattern should preserve that scope explicitly — document that the `state` dict is fresh per call, never reused across generations.

## Prior art

Searched open + closed issues/PRs for `guidance`, `registry`, `plugin`, `pluggable`, `hook`. Nothing matched this architectural proposal. Closest related work: **#703** (added MLX CFG/APG support — adds a variant, doesn't propose pluggability), **#832** (ADG batch bug fix), **#1000/#998** (DRY consolidation — supportive precedent). If I missed a prior discussion, happy to close this as duplicate and move the conversation there.

## The ask

**Are you open in principle to this pattern?** I'd rather get a read from you before investing in a draft PR.

- **Yes → I open a Draft PR** (~200–300 lines, PyTorch-only) against `main` for design review. We iterate on signatures / naming / docs together.
- **No / not yet → I drop the proposal** and continue contributing incremental HTTP-exposure PRs per variant (#1092 pattern). That's the status quo and I can live with it.
- **Maybe, but... → your feedback becomes the spec.** Anything I have wrong above, any constraint I'm not seeing, please push back.

Happy to refine the design based on whatever you say.

FlexOr2:
## Summary

Replaces the hardcoded `apg_forward` / `adg_forward` / `cfg_forward` dispatch inside every `generate_audio()` with a pluggable registry keyed by a string `guidance_variant` and configured by a free-form `guidance_params` dict. Exposes both on `/release_task`. API docs updated (en/ja/ko/zh).

Closes #1124. Maintainer approval quote (@ChuxiJ, 2026-04-22):
> approve, we can have a refact pr

Branched on top of `upstream/main` at `1d9d2d3`.

## Related PRs

- **#1123 (Draft — add `eta` and `momentum` to APG guidance)** — this refactor supersedes it. `eta` and `momentum` land as registered params on the `apg_classic` wrapper (pass via `guidance_params={"eta": 1.0, "momentum": -0.75}`). Happy to close #1123 once this lands.
- **#1091, #1092** — precursor work. No conflict: `use_adg=True` with the default `guidance_variant="apg_classic"` transparently resolves to the `adg` registered variant, preserving existing behaviour byte-identically.

## Design decisions

- **`typing.Protocol` for the wrapper signature, not Pydantic or abc.ABC.** Guidance runs in the per-step sampler loop on GPU — a Pydantic validator per call would add measurable overhead. The wrappers obey a structural `Protocol`: `fn(pred_cond, pred_uncond, guidance_scale, state, **params) -> Tensor`. This matches the Diffusers `callback_on_step_end` style.
- **Registry lives in its own module** (`acestep/models/common/guidance_registry.py`) rather than being folded into `apg_guidance.py`. Keeps the raw math functions as a thin numerical layer; the wrappers are the adapter. `apg_guidance.py` is untouched.
- **Unknown variant raises `ValueError`** listing the registered names. No silent default to APG — silent fallbacks make parameter bugs invisible in a diffusion stack.
- **State dict is the sampler's responsibility.** Sampler creates an empty dict at the start of `generate_audio`, stashes per-step values (`latents`, `sigma`, `step_role`) into it before each call, and drops it on return. Wrappers persist cross-step bookkeeping (APG's `MomentumBuffer`) in the same dict. `state["step_role"]` ∈ `{"main", "corrector"}` so Heun correctors can degrade APG → CFG and ADG → CFG-at-sigma=0 exactly as the pre-refactor code did.
- **CFG=1 is a no-op** by construction (`uncond + 1 * (cond - uncond) == cond`); no special case is needed. Documented in `guidance_cfg` docstring.

## Scope

**In:**
- PyTorch guidance dispatch for all four model families (`base`, `sft`, `xl_base`, `xl_sft`), both Euler and Heun samplers, main + corrector steps.
- HTTP surface: `guidance_variant` and `guidance_params` on `GenerateMusicRequest`, with snake_case + camelCase alias resolution, JSON-string decoding with `None` fallback on malformed input.
- End-to-end plumbing: HTTP body → `GenerationParams` → `generate_music` → `service_generate` → `_build_service_generate_kwargs` → `model.generate_audio`.
- Documentation: `docs/{en,ja,ko,zh}/API.md` gain rows for `guidance_variant` and `guidance_params` next to `use_adg`.

**Out:**
- **MLX.** `_mlx_apg_forward` in `acestep/models/mlx/dit_generate.py` uses a different state-dict shape and does not expose `eta`. MLX parity with the PyTorch registry is a separate patch; called out explicitly so nobody assumes they get free MLX support from this PR.

## Backward compatibility

The hard invariant: `generate_audio(..., guidance_variant="apg_classic", guidance_params={})` produces output bit-identical to the pre-refactor call site (`torch.allclose` at default tolerance `rtol=1e-5, atol=1e-8`, which on same-input deterministic reruns is effectively byte-identical). The legacy `use_adg=True` path also preserves its behaviour byte-identically by auto-resolving to the `adg` registered variant when the caller has not specified a non-default `guidance_variant`.

Proof (all passing):
- `guidance_registry_test::test_apg_classic_wrapper_matches_apg_forward` — wrapper vs raw `apg_forward` on fresh momentum.
- `guidance_registry_sampler_regression_test::test_full_euler_sequence_matches_hand_written_loop` — N-step Euler parity with momentum carry-over.
- `guidance_registry_sampler_regression_test::test_heun_sequence_matches_hand_written_loop` — Heun predictor+corrector parity for APG (predictor via `apg_forward`, corrector via `cfg_forward`).
- `guidance_registry_sampler_regression_test::test_heun_sequence_matches_hand_written_loop_with_zero_sigma_corrector` — Heun parity for ADG including the `sigma == 0` corrector safeguard.

## Test plan

- [x] `python -m unittest acestep.models.common.guidance_registry_test acestep.models.common.guidance_registry_sampler_regression_test` — 16 tests, all green.
- [x] `python -m unittest discover -s acestep -p "*_test.py" -t .` — matches `origin/main` (same pre-existing errors, no new failures). **+24 new tests** added by this PR vs the baseline.
- [x] Registry unit tests: default contents, decorator wiring, unknown-name `ValueError`, per-wrapper parity with raw functions, momentum state persistence, fresh-state semantics, corrector-step degradation.
- [x] Sampler-level regression tests (3 tests): Euler APG parity, Heun APG parity, Heun ADG parity with `sigma == 0` safeguard.
- [x] HTTP-level tests: Pydantic defaults, snake/camelCase alias resolution, JSON-string decoding with None fallback on malformed input, `GenerationParams` forwarding with and without the fields present on the request.
- [ ] Manual smoke: real generation with each of `{apg_classic, cfg, adg, adg_w_norm, adg_wo_clip}`. Hardware-bound, leaving to reviewer / CI with weights.

## Happy to adjust

Any decision flagged above is a starting point, not a commitment. If you prefer a different protocol signature, registry location, state-dict convention, or naming scheme — push back and I'll revise. The goal is to land something maintainable, not to litigate my initial choices.

## Summary by CodeRabbit

* **New Features**
  * Added pluggable guidance controls for music generation, including selectable variants and per-variant parameters.
  * Added model-aware defaults for DCW behavior and expanded support for direct-conditioning tasks.

* **Bug Fixes**
  * Improved handling of guidance parameters supplied as JSON, including invalid or unsupported values.

* **Tests**
  * Added coverage for guidance selection, aliases, parameter parsing, and end-to-end propagation.

* **Documentation**
  * Updated API documentation with guidance options, aliases, multipart JSON handling, and repainting controls.


FlexOr2:
@coderabbitai review

FlexOr2:
Thanks @coderabbitai — review addressed in `b80d91a`. Summary:

**Major (findings 1–5, the `use_adg` override bug):**
Core fix applied. `guidance_variant` default changed from `"apg_classic"` to `Optional[str] = None` across the HTTP Pydantic model, the request builder, `GenerationParams`, all four handler layers, and all four model `generate_audio` signatures. Resolution logic now distinguishes "caller omitted" (None → legacy `use_adg` fallback applies) from "caller explicitly set `apg_classic`" (always honored, never silently switched to `adg`). Default-value test expectations updated.

**Minor:**
- **Finding 6** (Ruff B905): `strict=True` added to all five `zip()` calls in the sampler regression tests.
- **Findings 8, 9** (docs): default column now shows `null` in all four language `API.md` files with an inline note on the `use_adg` fallback. Added multipart form-encoding guidance for `guidance_params`, camelCase alias mention (`guidanceVariant` / `guidanceParams`), and the confusing `{"eta": 1.0}` example shrunk to `{"eta": 0.1}`.

**Finding 7** (`3.14` vs `math.pi` in `angle_clip` default): verified the registry wrappers' defaults (`3.14 / 6` for `adg`, `3.14 / 3` for `adg_w_norm`) match the pre-refactor values in `apg_guidance.py` exactly. Consistent by design; no change needed.

PR still Draft — happy to take another round of feedback.

FlexOr2:
Second round addressed in `8ecd184`:

- **Docstrings** on the two new `GenerationParams` fields (slotted into the existing Attributes block in `inference.py`) and on `generate_audio` in all four model variants (`base`, `sft`, `xl_base`, `xl_sft`) — covers `guidance_variant`, `guidance_params`, and the `ValueError` path from unknown variant names.
- **Narrowed `except Exception` → `except json.JSONDecodeError`** around `guidance_params` JSON parsing in `release_task_request_builder.py`; import promoted to the stdlib group at module top per PEP 8.

No behaviour change, tests unchanged.

FlexOr2:
@coderabbitai review

FlexOr2:
Rebased onto upstream main at 14c0211. Range-diff confirms all six existing commits are semantically unchanged. The focused test run exposed the same stale upstream RequestParser test stub seen in #1092; follow-up fb611b7 updates the shared stub and removes the redundant guidance-only subclass. Verification: 44 passed plus 4 subtests; focused Ruff checks for the API, handler, and new registry modules are clean. New fork head: fb611b7.

PR #1130, opened 2026-04-22
Security: API key accepted from request body (`ai_token`) can leak via logs and intermediaries

tuanaiseo:
## Problem

The authentication helper accepts credentials from `body['ai_token']` in addition to the `Authorization` header. Tokens in JSON bodies are more likely to be logged by application middleware, error handlers, or upstream gateways, and are not handled by standard auth-header redaction rules. This increases the risk of secret disclosure.

**Severity**: `medium`
**File**: `acestep/api/http/auth.py`

## Solution

Prefer `Authorization: Bearer <token>` exclusively for API authentication. If backward compatibility is required, gate body-token support behind an explicit legacy flag and ensure body fields containing secrets are redacted from logs.

## Changes

- `acestep/api/http/auth.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

## Summary by CodeRabbit

* **Breaking Changes**
  * Authentication now requires credentials exclusively via the `Authorization` header; the alternative request body method is no longer supported.
  * Updated error messaging to reflect that only the `Authorization` header is required for authentication.

### DIFF
--- a/acestep/api/http/auth.py
+++ b/acestep/api/http/auth.py
@@ -23,7 +23,7 @@ def set_api_key(key: Optional[str]) -> None:
 def verify_token_from_request(
     body: dict[str, Any], authorization: Optional[str] = None
 ) -> Optional[str]:
-    """Validate request auth from body ``ai_token`` or Authorization header.
+    """Validate request auth from Authorization header.
 
     Args:
         body: Parsed request payload dictionary.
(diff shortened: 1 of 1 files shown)


ChuxiJ (changes requested):
Thanks for the security cleanup. I agree with the direction of removing body-based credentials, but I don't think this is ready to merge yet.

Blocking issues:

1. The existing focused auth tests fail on this branch:

```text
PYTHONPATH=/private/tmp/ace-pr1130.M9Nbam /Users/gongjunmin/timedomain/ACE-Step-1.5/.venv/bin/python -m unittest acestep.api.http.auth_test
Ran 6 tests
FAILED (failures=2, errors=1)
```

The failures are expected for this behavioral change, but the PR needs to update the tests to describe the new contract instead of leaving the old body-token expectations broken.

2. This still accepts a raw API key in the Authorization header. For a security PR that removes `ai_token` from the body, we should also tighten this path to require `Authorization: Bearer <token>` consistently in both `verify_token_from_request` and `verify_api_key`; otherwise credentials can still be sent through a non-standard header value that is easier to mishandle.

3. Please update the API docs that currently advertise `ai_token` in the request body, and add regression tests covering:
- body `ai_token` is ignored/rejected;
- `Authorization: Bearer <token>` succeeds;
- raw `Authorization: <token>` is rejected;
- missing and invalid header behavior.

Once those are addressed, this should be a good high-priority security fix to land.

PR #1131, opened 2026-04-22
Security: Authentication is fail-open when API key is unset

tuanaiseo:
## Problem

Both `verify_token_from_request` and `verify_api_key` return success when `_api_key` is `None`. If deployment/configuration accidentally omits setting the key, all protected endpoints become publicly accessible without authentication.

**Severity**: `medium`
**File**: `acestep/api/http/auth.py`

## Solution

Adopt fail-closed behavior by default: require explicit configuration to disable auth (e.g., `ALLOW_UNAUTHENTICATED=false` by default), and log a high-severity startup warning/error when auth is disabled.

## Changes

- `acestep/api/http/auth.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API key validation error handling. When the API key is not configured, the system now returns a proper error response instead of proceeding with incomplete authentication, making configuration issues easier to diagnose.

### DIFF
--- a/acestep/api/http/auth.py
+++ b/acestep/api/http/auth.py
@@ -13,7 +13,7 @@ def set_api_key(key: Optional[str]) -> None:
     """Set the process-level API key used by auth verification helpers.
 
     Args:
-        key: API key string, or ``None`` to disable auth enforcement.
+        key: API key string, or ``None`` when auth is unconfigured.
     """
 
     global _api_key
@@ -30,14 +30,14 @@ def verify_token_from_request(
         authorization: Optional ``Authorization`` header value.
 
     Returns:
-        Validated token value, or ``None`` when auth is disabled.
+        Validated token value.
 
     Raises:
         HTTPException: If auth is required and token is missing/invalid.
     """
 
     if _api_key is None:
-        return None
+        raise HTTPException(status_code=503, detail="API key is not configured")
 
     ai_token = body.get("ai_token") if body else None
     if ai_token:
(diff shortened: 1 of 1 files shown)


ChuxiJ (changes requested):
Thanks for raising this. The fail-open behavior is a real security concern, but I don't think this exact change is safe to merge as-is because it changes the existing no-auth/default mode without an explicit compatibility story.

Blocking issues:

1. The focused auth tests fail on this branch:

```text
PYTHONPATH=/private/tmp/ace-pr1131.u9k0ii /Users/gongjunmin/timedomain/ACE-Step-1.5/.venv/bin/python -m unittest acestep.api.http.auth_test
Ran 6 tests
FAILED (errors=1)
```

The failing test is the current "auth disabled when no API key is configured" path. If we intentionally change that contract, the PR should update the tests and document the new startup/auth mode clearly.

2. Please avoid making `_api_key is None` unconditionally fail every request. We need an explicit policy boundary, for example:
- secure/fail-closed mode when unauthenticated API access is not explicitly allowed;
- an explicit opt-out flag for local/dev unauthenticated mode;
- startup logging that makes the chosen mode obvious.

3. Please keep behavior aligned between the HTTP API auth helpers and the Gradio API auth helpers, or explicitly split/document why they differ.

With that compatibility and test coverage in place, this is worth landing as a high-priority hardening change.

Issue #426, opened 2026-02-10
Only using 1 of multiple GPUs available


KillerX629:
The program only defaults to using torch.device(0) at startup, calculating available VRAM from that line, however, if multiple GPUs are available, they go unused.

listing the devices discoverable by torch and fitting models appropiately should be an option, currently trying to fit sft and 4b in 24Gb fails.

Petruha04:
ComfyUI?

KillerX629:
> ComfyUI? 

Nope  the gradio ui included to test

KillerX629:
Here's the issue,I THINK.
When reading memory for configuring the GPUs, only the first device is evaluated (cuda:0)

```python

    try:
        import torch
        if torch.cuda.is_available():
            # Get total memory of the first GPU in GB
            total_memory = torch.cuda.get_device_properties(0).total_memory
            memory_gb = total_memory / (1024**3)  # Convert bytes to GB
            device_name = torch.cuda.get_device_name(0)
            is_rocm = hasattr(torch.version, 'hip') and torch.version.hip is not None
            if is_rocm:
                logger.info(f"ROCm GPU detected: {device_name} ({memory_gb:.1f} GB, HIP {torch.version.hip})")
```
    

Here and on inference, it should be a sum of all cuda devices available, no?

alfrentgen:
The gradio does not support multigpu inference actually. Looking at the code, everything is done an cuda:0 device. It would be a great thing to implement multi-GPU model offloading, i.e. lm model goes to cuda:1, evertything else goes to cuda:0.
Or shard model via pytorch FSDP2 over several GPUs.

It would definitely a big step towards users who have enough VRAM but on a few GPUs.

ChuxiJ:
Thank you for raising this! Multi-GPU support is indeed a valuable feature request.

### Current Behavior

You're correct - ACE-Step currently defaults to `torch.device(0)` and calculates available VRAM from that single GPU. The Gradio UI doesn't support distributing models across multiple GPUs.

### Current Workarounds

**1. Manual device specification via API**

If you're using the API directly (not Gradio UI), you can manually specify device placement in code:
```python
# Place DiT on GPU 0, VAE on GPU 1
dit_model = dit_model.to('cuda:0')
vae = vae.to('cuda:1')
```

**2. Use CPU offloading**

For large models that don't fit in 24GB, try:
```bash
uv run acestep --offload_to_cpu --offload_dit_to_cpu
```
[...]

PR #1149, opened 2026-04-26
issue-fix-426: Multi-GPU support by component offloading

imsarang:
Multi GPU support for component offloading.
issue link: https://github.com/ace-step/ACE-Step-1.5/issues/426

Earlier: GPU 0 used to perform dit, vae and lm tasks.

Now:
expected result GPU i will take respective component.
GPU allocation is done by ranking at runtime, where one gpu has one component to execute.

## Summary by CodeRabbit

* **New Features**
  * Per-component GPU mapping auto-selects, validates and is applied at startup and service initialization; environment overrides still honored.
  * LLM and model components can be targeted to different GPUs via the resolved mapping.

* **UI**
  * GPU info now shows a per-component GPU hint (DiT/VAE/LM) when applicable.

* **Tests**
  * Added/updated tests for device resolution, validation, multi-/single-GPU behavior, no-CUDA handling, and LLM device selection.


pokepress:
So, I tried using this by creating a branch off the current code in the main branch and merging yours in. No issues getting it to start up, but I didn't see it move anything to the second GPU. For the record, my setup is:

ThinkStation p720 with 32GB RAM on Windows 10
GPU 1: 4060ti 16GB
GPU 2: 3060 12GB

The generated hint was: DiT=cuda:0, VAE=cuda:1, LM=cuda:1

Suggesting the DiT music model would remain on the 4060ti, and the others would move onto 3060. However, only the 4060 ti was used, and the VAE and LM were offloaded to the CPU. Do I need to override the setup based on my total memory to make this work?

imsarang:
according to the code, all the available gpus should be shown and ranked, based on the ranking the module should be offloaded.
You can try by overriding the setup, but i dont think this will be the permanent fix.
Possibly a static mapping might get involved, but we need a better solution, I will push my changes based on this.

Thanks for the feedback, will update the code.

pokepress:
> according to the code, all the available gpus should be shown and ranked, based on the ranking the module should be offloaded. You can try by overriding the setup, but i dont think this will be the permanent fix. Possibly a static mapping might get involved, but we need a better solution, I will push my changes based on this.
> 
> Thanks for the feedback, will update the code.

I believe I also had "offload to CPU" on, which is checked by default in this configuration. I did turn off dit to CPU, though.

pokepress:
So, I gave editing the code a try myself, and wanted to explain some of the things that make this more difficult than it might seem:

1. The existing code has a parameter that allows you to specify a specific device for the (L)LM. However, many places in code (both in inference and the general gpu utils) actually just hard code an index of 0.
2. Some supporting frameworks (in particular, VLLM) don't make it easy to specify which device a model should be sent to.
3. If models still need to be offloaded to the CPU, that needs to take into account the model may be on a device other than 0
4. Some cuda operations require the relevant tensors to be on the same device. I don't know if this can be reworked.

PR #1190, opened 2026-05-06
fix(jetson): remove GPL video codecs, fix ffmpeg/examples/healthcheck

toolboc:
Addresses review feedback from #793 and #735:

- Remove unused x264/x265 GPL video codecs from FFmpeg build (audio-only paths don't need them, avoids GPL license complexity)
- Remove `--disable-programs` so `ffmpeg` CLI binary is available for MP3 export in `audio_utils.py`
- Pin `bitsandbytes>=0.49.0,<0.51` for aarch64 stability
- Fix healthcheck port variable to `API_PORT` matching docker-compose.jetson.yml
- Add examples symlink workaround for `_get_project_root()` path resolution bug in `api_routes.py`

## Summary by CodeRabbit

* **Compatibility**
  * Updated Jetson media support to focus on MP3, Opus, and Vorbis formats.
  * Restricted the supported `bitsandbytes` version range for improved compatibility.

* **Improvements**
  * Healthchecks now verify both Gradio and API endpoints with configurable port fallbacks.
  * Runtime setup supports additional example paths and explicit API port configuration.
  * Added installation verification for nano-vLLM.

* **Documentation**
  * Documented supported model configurations and approximate VRAM requirements for Jetson deployments.

### DIFF
--- a/Dockerfile.jetson
+++ b/Dockerfile.jetson
@@ -81,8 +81,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         # FFmpeg build dependencies (we build FFmpeg 7 from source for torchcodec)
         nasm \
         yasm \
-        libx264-dev \
-        libx265-dev \
         libmp3lame-dev \
         libopus-dev \
         libvorbis-dev \
@@ -105,14 +103,10 @@ RUN cd /tmp \
         --prefix=/usr/local \
         --enable-shared \
         --disable-static \
-        --enable-gpl \
-        --enable-libx264 \
-        --enable-libx265 \
         --enable-libmp3lame \
         --enable-libopus \
         --enable-libvorbis \
         --disable-doc \
-        --disable-programs \
     && make -j$(nproc) \
     && make install \
     && ldconfig \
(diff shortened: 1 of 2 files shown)


ChuxiJ (changes requested):
Thanks for continuing to clean up the Jetson image. I am not merging this yet because the remaining compose guidance can still point low-VRAM Jetson users into a non-starting configuration.

The current `docker-compose.jetson.yml` still presents `acestep-5Hz-lm-0.6B` as suitable for 4GB-class devices. Per the latest review note, please remove the 4GB Orin Nano recommendation for that LM, limit it to 8GB-class devices, or explicitly state that LM initialization must stay disabled on systems with 6GB VRAM or less.

Once that guidance is corrected, this Docker-only PR should be easier to approve. I did not run a full Jetson image build locally in this environment.

toolboc:
Addressed the low-VRAM guidance in 98fe6c4: the 0.6B LM recommendation is now limited to Orin Nano 8GB, and devices with 6GB VRAM or less are instructed to set `ACESTEP_INIT_SERVICE=false` and initialize only the DiT. `docker compose -f docker-compose.jetson.yml config -q` and `git diff --check` both pass.

ChuxiJ (approved):
Thanks for the follow-up. The low-VRAM Jetson guidance blocker from my previous review is addressed in `148567f`: 0.6B LM is now limited to Orin Nano 8GB, and <=6GB devices are told to keep LM init disabled and initialize only the DiT.

I also checked:

```bash
git diff --check origin/main...HEAD
```

No whitespace/shell-continuation issues were reported. I could not use this local machine's `docker compose config -q` as a PR signal because Docker Compose v2.19.1 rejects the existing `env_file.required` syntax on clean `main` too, so that failure is a baseline/tooling-version issue here rather than a regression from this PR.

I did not run a full Jetson image build in this environment, but the remaining diff is Docker/compose scoped and the prior blocker is closed. Approving.

PR #1194, opened 2026-05-09
fix: stale DiT instruction after Simple Mode sample creation + add setuptools dep

EmilioPlumed:
To reproduce: Start the UI, switch to Remix mode (which sets instruction to the LM/cover value), then switch to Simple mode, create a sample, and click Generate. Without this fix, instruction_display_gen remains stale and the DiT receives cover-mode conditioning, producing white noise.

Two related fixes:

1. **Stale instruction_display_gen in Simple Mode → Generate Full Song** When create_sample_btn fires it sets generation_mode="Custom" (task_type="text2music"). If task_type was already "text2music" the task_type.change event never fires, leaving instruction_display_gen holding a stale value — typically the LM instruction "Generate audio semantic tokens based on the given conditions:" from a prior Remix/Cover session. The DiT then receives cover-mode text conditioning while operating in text2music mode, which produces near-silent latents (peak ~0.003) that normalization amplifies into audible white noise.

   Fix: chain a .then() on create_sample_btn.click that unconditionally resets instruction_display_gen to TASK_INSTRUCTIONS["text2music"] after the sample is created, regardless of prior mode state.

2. **pytorch_wavelets fails to import on Python 3.12+ (pkg_resources missing)** pytorch_wavelets' DTCWT coefficient loader uses pkg_resources from setuptools, which is no longer implicitly available in Python 3.12+ environments (e.g. uv-managed venvs). This causes the DCW wavelet correction to silently fall back to a no-op even when enabled.

   Fix: add setuptools as an explicit dependency in pyproject.toml so it is always present alongside pytorch-wavelets.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure the instruction display refreshes immediately after creating a sample for text2music, so the UI shows current generation parameters reliably.

* **Chores**
  * Pinned build dependency: setuptools version restricted to <72.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ace-step/ACE-Step-1.5/pull/1194)

### DIFF
--- a/acestep/ui/gradio/events/wiring/generation_mode_wiring.py
+++ b/acestep/ui/gradio/events/wiring/generation_mode_wiring.py
@@ -10,6 +10,7 @@ import gradio as gr
 
 from .. import generation_handlers as gen_h
 from .context import GenerationWiringContext
+from acestep.constants import MODE_TO_TASK_TYPE
 
 
 def _on_repaint_mode_change(mode, current_strength, memory):
@@ -44,6 +45,7 @@ def register_generation_mode_handlers(
 
     generation_section = context.generation_section
     results_section = context.results_section
+    dit_handler = context.dit_handler
     llm_handler = context.llm_handler
 
     # Shared handler for mode-change and initial page load — extracted to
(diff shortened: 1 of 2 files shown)


EmilioPlumed:
Sonnet told me this: setuptools<72 is intentional. pytorch-wavelets (already a declared dependency) imports pkg_resources at module load time, which was removed in setuptools 72+. Without this pin, DCW silently falls back to a no-op on Python 3.12+ uv-managed environments, causing audio quality degradation that is difficult to diagnose. There is no cleaner place to express this constraint given that uv.lock is gitignored in this repo.

Issue #665, opened 2026-02-21
ROCm not working in repaint: No soundfile backend in process_reference_audio


shipped-it:
**Describe the bug**
torchaudio relies on torchcodec, so we can't use an AMD GPU for repainting.
```
2026-02-21 20:25:02.418 | ERROR    | acestep.core.generation.handler.generate_music:generate_music:183 - [generate_music] Generation failed
Traceback (most recent call last):

  File "/mnt/HeavyData/AI/ACE-Step-1.5/.venv/lib/python3.11/site-packages/torchaudio/_torchcodec.py", line 82, in load_with_torchcodec
    from torchcodec.decoders import AudioDecoder

ModuleNotFoundError: No module named 'torchcodec'


The above exception was the direct cause of the following exception:


Traceback (most recent call last):

  File "/home/user/.local/share/uv/python/cpython-3.11.14-linux-x86_64-gnu/lib/python3.11/threading.py", line 1002, in _bootstrap
    self._bootstrap_inner()
    │    └ <function Thread._bootstrap_inner at 0x7f1710ee5c60>
    └ <WorkerThread(AnyIO worker thread, started daemon 139710781454016)>
  File "/home/user/.local/share/uv/python/cpython-3.11.14-linux-x86_64-gnu/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
    self.run()
    │    └ <function WorkerThread.run at 0x7f11d5706a20>
    └ <WorkerThread(AnyIO worker thread, started daemon 139710781454016)>
  File "/mnt/HeavyData/AI/ACE-Step-1.5/.venv/lib/python3.11/site-packages/anyio/_backends/_asyncio.py", line 986, in run
    result = context.run(func, *args)
             │       │   │      └ (<generator object register_generation_run_handlers.<locals>.generation_wrapper at 0x7f11d8ba1000>,)
             │       │   └ <function run_sync_iterator_async at 0x7f1302f76d40>
             │       └ <method 'run' of '_contextvars.Context' objects>
             └ <_contextvars.Context object at 0x7f11dd1fb280>
  File "/mnt/HeavyData/AI/ACE-Step-1.5/.venv/lib/python3.11/site-packages/gradio/utils.py", line 833, in run_sync_iterator_async
    return next(iterator)
                └ <generator object register_generation_run_handlers.<locals>.generation_wrapper at 0x7f11d8ba1000>
[...]

PR #1242, opened 2026-06-12
Fixed 'torchcodec' not supported issue on XPU and ROCm when calling `torchaudio.save`

fisheryv:
In torchaudio 2.9+ and later versions, `torchaudio.save()` internally calls `save_with_torchcodec`, even if the `backend='soundfile'` or `backend='ffmpeg'` parameter is specified, which will be ignored. See [https://docs.pytorch.org/audio/stable/generated/torchaudio.save.html#torchaudio.save](https://docs.pytorch.org/audio/stable/generated/torchaudio.save.html#torchaudio.save). When `torchcodec` is not installed (e.g., XPU and ROCm do not support `torchcodec`), this results in an `ImportError: TorchCodec is required for save_with_torchcodec` or `ModuleNotFoundError: No module named 'torchcodec'`.

I implemented a dual-path compatibility scheme with `torchcodec` as the priority and `soundfile/ffmpeg` as fallbacks, to address the issue where XPU and ROCm do not support `torchcodec`, or `torchcodec` incompatibility prevents saving generated audio.

Related issues: #665

## Summary by CodeRabbit

* **New Features**
  * Added optional topology correction during audio generation, with validation to ensure only valid results are applied.
  * Improved compatibility with multiple audio-saving backends and formats.

* **Bug Fixes**
  * Audio exports now recover more reliably when the primary saving method is unavailable or fails.
  * Improved WAV32 handling and fallback support for Opus, AAC, FLAC, and WAV files.
  * Hardened audio export in the flow-edit smoke test.

* **Tests**
  * Expanded coverage for backend detection, format routing, codec selection, and fallback behavior.


mnj:
This solved the issue here, where I otherwise had zero progress on getting it working:

Arch Linux / 20GB AMD 7900 XT

Followed https://github.com/ace-step/ACE-Step-1.5/blob/main/docs/en/ACE-Step1.5-Rocm-Manual-Linux.md
But used rocm 7.2 version of pytorch
This patch to get it to generate the mp3s.

ChuxiJ (changes requested):
Thanks for the ROCm/XPU fix, and the user confirmation from Arch/7900 XT is useful. I cannot merge this yet because the focused audio utility suite is not green on the current PR head `a877d05`.

```bash
uv run --no-sync python -m unittest acestep.audio_utils_test
```

Result: 50 tests ran, 3 failures and 2 errors. The failing coverage is all around AAC/Opus dispatch expecting the ffmpeg path:

- `AudioSaverFormatTests.test_save_audio_aac_uses_ffmpeg_backend`
- `AudioSaverFormatTests.test_save_audio_opus_uses_ffmpeg_backend`
- `AudioSaverFormatTests.test_numpy_array_input_with_opus`
- `AudioSaverFormatTests.test_save_audio_validates_aac_format`
- `AudioSaverFormatTests.test_save_audio_validates_opus_format`

Also, the current `_check_torchcodec_available()` still opens a `NamedTemporaryFile(delete=True)` and calls `torchaudio.save()` while the file handle is open. That keeps the earlier Windows-specific import-time risk alive. Please close the temp file before probing, clean it up in a `finally`, and broaden the probe exception handling so importing `acestep.audio_utils` cannot fail on platform-specific save/probe errors.

Once the test expectations match the intended backend routing and the probe is hardened, this should be worth another pass because it addresses a real ROCm/XPU pain point.

ChuxiJ (changes requested):
Thanks for updating this, but the current head `79da6c4` is still not merge-ready.

I re-ran the focused audio utility suite:

```bash
python -m unittest acestep.audio_utils_test
```

Result: 50 tests ran, 3 failures and 2 errors. The failing cases are still the AAC/Opus routing assertions where the tests expect the ffmpeg path but `_save_via_ffmpeg` is not called.

There is also a new functional blocker outside the original audio-export scope: `acestep/core/generation/handler/service_generate_execute.py` now passes `dcw_enabled=dcw_enabled` on the MLX path, but `dcw_enabled` is not defined in that local scope. That can raise `NameError` before `_mlx_run_diffusion()` executes and force/fail the backend path.

Please keep this PR focused on the torchcodec/audio-export compatibility fix, or split the topology-corrector work into a separate PR. For this PR, the minimum merge bar is:

- `acestep.audio_utils_test` green;
- no undefined `dcw_enabled` reference on the MLX path;
- torchcodec probe hardened as discussed earlier.

PR #1245, opened 2026-06-15
fix(training_v2): respect --precision fp32 on MPS

### LINKED ISSUE #1244: training_v2 ignores --precision fp32 when selecting Fabric precision on MPS
## Summary

`train_fixed --device mps --precision fp32` displays/loads fp32 but the training loop still selects MPS defaults (`torch.float16`, `16-mixed`), so the user cannot force true fp32 training from the CLI.

## Expected

`--precision fp32` should select `torch.float32`, Fabric `32-true`, and avoid autocast.

## Actual

`FixedLoRATrainer` calls `_select_compute_dtype(device_type)` and `_select_fabric_precision(device_type)` without passing `cfg.precision`, so MPS always resolves to fp16/`16-mixed`.

A focused local check against the current `HEAD` sources shows:

```text
HEAD _select_compute_dtype signature: (device_type: str) -> torch.dtype
HEAD _select_fabric_precision signature: (device_type: str) -> str
HEAD _select_compute_dtype("mps"): torch.float16
HEAD _select_fabric_precision("mps"): 16-mixed
trainer passes cfg.precision to dtype helper: False
trainer passes cfg.precision to fabric helper: False
```

## Impact

Makes fp32 an ineffective escape hatch for MPS numerical instability / AMP issues in Side-Step `training_v2`.

## Related

Related MPS training context, but not duplicates:

- #367: MPS mixed precision / FP16 gradients, closed unmerged
- #401: MPS training stabilization, merged
- #619: MPS LoRA training issues, closed

## Proposed fix

Thread requested precision through the helper functions and skip autocast when dtype is fp32.

denisroldan:
Fixes #1244

## Summary

- Threads requested CLI precision into `training_v2` compute/Fabric precision selection.
- Makes `--precision fp32` use `torch.float32` and Fabric `32-true`.
- Skips autocast when dtype is fp32.

## Why

Previously MPS always resolved to fp16/`16-mixed` inside training even when users passed `--precision fp32`. The trainer also selected a dtype before constructing `FixedLoRAModule`, but the module recomputed the device default instead of keeping the selected dtype.

## Tests

- PASS: `.venv/bin/python -m py_compile acestep/training_v2/fixed_lora_module.py acestep/training_v2/trainer_fixed.py`
- PASS: `.venv/bin/python -m unittest acestep.training_v2.fixed_lora_module_precision_test` (`Ran 8 tests`)

## Notes

- Does not change `auto` defaults.
- Does not attempt to solve all MPS non-finite-loss cases; this only makes the documented fp32 escape hatch actually work.

## Release Notes

* **New Features**
  * Training now supports configurable precision modes (`auto`, `fp32`, `fp16`, `bf16`) with device-aware defaults.

* **Bug Fixes**
  * Mixed-precision autocasting is now enabled only when using CUDA/XPU/MPS *and* when the selected precision is not `fp32`.

* **Tests**
  * Added unit tests covering precision selection mappings, invalid precision errors, `FixedLoRAModule` dtype preservation, autocast behavior for `fp32`, and trainer wiring to pass configured precision through.

### DIFF
--- a/acestep/training_v2/fixed_lora_module.py
+++ b/acestep/training_v2/fixed_lora_module.py
@@ -80,7 +80,15 @@ def _normalize_device_type(device: Any) -> str:
     return str(device)
 
 
-def _select_compute_dtype(device_type: str) -> torch.dtype:
+def _select_compute_dtype(device_type: str, requested_precision: str = "auto") -> torch.dtype:
+    """Select tensor compute dtype from requested precision or device defaults."""
+    requested_precision = requested_precision or "auto"
+    if requested_precision == "fp32":
+        return torch.float32
+    if requested_precision == "fp16":
+        return torch.float16
+    if requested_precision == "bf16":
+        return torch.bfloat16
     if device_type in ("cuda", "xpu"):
         return torch.bfloat16
     if device_type == "mps":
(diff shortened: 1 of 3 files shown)


ChuxiJ (approved):
Reviewed as P0 triage: scoped MPS fp32 precision fix with focused unit coverage.

PR #1267, opened 2026-07-15
fix(api): use per-run seeds in sequential MPS cover mode

gonasooc:
## Problem

On MPS, cover-family tasks (`cover`, `cover-nofsq`) with `batch_size > 1` are split into sequential batch-1 runs by `run_generation_with_optional_sequential_cover_mode`. Each sequential run, however, receives the **full** requested seed list, so downstream seed preparation picks the first seed every time. All "variants" of the batch are generated with the same seed and come out byte-identical.

The batched path (CUDA and other devices that don't enter sequential mode) consumes one seed per batch item correctly, so this only affects the MPS sequential branch.

## Reproduction

- Device: MPS (Apple Silicon)
- Request: `task_type=cover`, `batch_size=2`, `use_random_seed=false`, `seeds=[20260713, 20260714]`
- Actual: both outputs are generated with seed `20260713` (identical audio files)
- Expected: run 1 uses `20260713`, run 2 uses `20260714` (matching batched behavior)

## Fix

Add `_sequential_seed_slices`, which splits the requested seeds into per-run slices before the sequential loop, and assign each run its own slice:

- explicit seed list → run *i* gets `[seeds[i]]`; extra seeds past `batch_size` are dropped, matching batched truncation
- a single int seed is treated as a one-element list
- more runs than seeds → extra runs get `None`, so seed preparation falls back to a random seed (matching batched behavior)
- `use_random_seed=true` or missing/empty seeds → config value passed through unchanged for every run

## Blast radius

- Product code touched: only the MPS sequential cover branch in `acestep/api/job_generation_runtime.py` (plus its unit test file); the helper is invoked exclusively inside the `handler_device == "mps"` block.
- **Non-target platforms unchanged** — code path is unreachable for CUDA/CPU/XPU (they never enter sequential mode).
- No behavior change for `use_random_seed=true` requests or for `batch_size == 1`.
- Additive change only (no lines removed).

## Regression checklist

- [x] `python -m unittest acestep.api.job_generation_runtime_test` — 8 tests OK (5 new + 3 pre-existing)
- [x] New tests cover: per-run seed consumption (`cover` and `cover-nofsq`), fallback past the seed list, extra seeds beyond `batch_size`, `use_random_seed` passthrough
- [x] Verified end-to-end on MPS (M-series): `batch_size=2` cover request with two fixed seeds now produces two distinct outputs; previously byte-identical
- [x] Based on current `main`; the fix applies to both `cover` and `cover-nofsq`

## Reviewer notes

- `config.seeds` is reassigned in place across sequential runs; this follows the pre-existing pattern in this function (`config.batch_size` is already mutated the same way), and `config` is rebuilt per job so nothing reads the original list afterwards. The per-audio seed echo in the response comes from each run's actual seed, so it stays accurate.
- Out of scope: seed handling on non-MPS paths and LM chunk seeding are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sequential cover generation on Apple silicon devices by ensuring each run uses its assigned seed.
  * Prevented missing seeds from causing inconsistent runs by safely filling unavailable values.
  * Preserved existing random-seed behavior and ignored excess seeds beyond the requested batch size.

* **Tests**
  * Added coverage for seeded, random-seeded, and no-frequency-separation cover generation scenarios.

### DIFF
--- a/acestep/api/job_generation_runtime.py
+++ b/acestep/api/job_generation_runtime.py
@@ -5,6 +5,22 @@ from __future__ import annotations
 from typing import Any, Callable
 
 
+def _sequential_seed_slices(config: Any, sequential_runs: int) -> list:
+    """Split requested seeds into per-run slices for sequential cover mode.
+
+    Runs beyond the provided seed list fall back to ``None`` so downstream
+    seed preparation draws a random seed, matching batched behavior.
+    """
+    if getattr(config, "use_random_seed", True):
+        return [getattr(config, "seeds", None)] * sequential_runs
+    seeds = getattr(config, "seeds", None)
+    if isinstance(seeds, int):
+        seeds = [seeds]
+    if not isinstance(seeds, list) or len(seeds) == 0:
+        return [seeds] * sequential_runs
+    return [[seeds[run_idx]] if run_idx < len(seeds) else None for run_idx in range(sequential_runs)]
+
+
 def run_generation_with_optional_sequential_cover_mode(
     *,
     req: Any,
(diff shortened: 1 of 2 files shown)


gonasooc:
Closing this for now — we'll keep the fix in our local checkout and may resubmit later.

PR #1275, opened 2026-07-27
fix(deps): declare python-multipart as a direct dependency

tsondo:
## Summary

The API server uses FastAPI `UploadFile`/`Form` endpoints (`acestep/api/route_setup.py`, `acestep/api/http/release_task_audio_paths.py`), which require `python-multipart` at route-definition time. It is currently only present transitively through `gradio`, so any install without gradio (e.g. a headless API-server deployment) fails at import with FastAPI's "Form data requires python-multipart" error.

This declares it explicitly in `pyproject.toml` and `requirements.txt`. No behavior change for existing installs — the package is already present in any environment that has gradio.

Complementary to #1276 (making `acestep.api_server` importable without gradio), which is the scenario that exposes this gap — but this fix stands on its own: load-bearing dependencies should be declared, not inherited.

## Summary by CodeRabbit

* **Chores**
  * Added support for multipart form data handling.
  * Updated the project’s runtime requirements to include the necessary multipart-processing package.


ChuxiJ:
Thanks for the clean dependency fix. I checked this locally in a temporary worktree and `uv lock --check` reports that `uv.lock` needs to be updated after adding `python-multipart>=0.0.18`.

Could you please include the corresponding `uv.lock` update in this PR? After that this should be straightforward to merge.

ChuxiJ (approved):
Verified the dependency declaration and lockfile consistency locally with uv lock --check. This is a small, direct fix for headless FastAPI installs where python-multipart is required at route definition time.

PR #1276, opened 2026-07-27
fix(api): make api_server importable without gradio installed

tsondo:
## Summary

`acestep.api_server` only needs one helper from the UI layer (`_build_generation_info`), but importing it via `acestep.ui.gradio.events.results_handlers` drags in the entire gradio UI package at import time — so the API server cannot start in environments where gradio is not installed (headless / API-only deployments that consume ACE-Step purely through REST).

Three changes keep the API path headless while leaving Gradio UI behavior unchanged when gradio is installed:

- **`acestep/ui/gradio/__init__.py`** — expose `create_gradio_interface` lazily (PEP 562 `__getattr__`) instead of importing the interfaces package eagerly
- **`acestep/ui/gradio/events/__init__.py`** — move the `.wiring` import (whose handler modules import gradio at module level) inside the two setup functions, which only run under the Gradio UI
- **`acestep/api_server.py`** — import `_build_generation_info` from its defining module (`events/results/generation_info.py`, already headless-safe with its own lazy gradio fallback) instead of the `results_handlers` facade

## Testing

- `import acestep.api_server` succeeds with gradio imports blocked (simulating an environment without gradio)
- `from acestep.ui.gradio import create_gradio_interface` and `from acestep.ui.gradio.events import setup_event_handlers, setup_training_event_handlers` still work with gradio installed
- `results_handlers_facade_test.py` and `i18n_thread_safety_test.py` pass

Related: #1275 declares `python-multipart` directly — headless installs need it since it currently only arrives via gradio.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved headless and API usage when Gradio is not installed.
  * Prevented unnecessary UI-related loading during standard application imports.
  * Preserved existing generation result handling and event registration behavior.

### DIFF
--- a/acestep/api_server.py
+++ b/acestep/api_server.py
@@ -95,7 +95,9 @@ from acestep.inference import (
     create_sample,
     format_sample,
 )
-from acestep.ui.gradio.events.results_handlers import _build_generation_info
+# Import from the defining module (not the results_handlers facade) so the API
+# server stays importable without gradio installed.
+from acestep.ui.gradio.events.results.generation_info import _build_generation_info
 
 def _get_project_root() -> str:
     current_file = os.path.abspath(__file__)
--- a/acestep/ui/gradio/__init__.py
+++ b/acestep/ui/gradio/__init__.py
@@ -5,6 +5,8 @@ submodules that do not need gradio itself (e.g. the headless helpers used by
 ``acestep.api_server``) does not require the gradio package to be installed.
 """
 
+__all__ = ["create_gradio_interface"]
+
 
 def __getattr__(name):
     if name == "create_gradio_interface":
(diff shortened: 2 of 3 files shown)


ChuxiJ:
This is the right direction for keeping `acestep.api_server` importable in headless installs. Before merge, could you please address the two small review items CodeRabbit left?

- expose `create_gradio_interface` through `__all__` while keeping the lazy `__getattr__` facade;
- expand `setup_training_event_handlers` docs so its inputs/side-effect return behavior are documented.

Once those are in, I can do a focused final pass and merge it.

ChuxiJ (approved):
Verified locally by importing acestep.api_server with gradio intentionally blocked. The lazy Gradio exports and direct generation_info import keep the API path usable for headless deployments, and the earlier review nits are addressed.

PR #1283, opened 2026-08-02
fix: make same-seed generations reproducible (LM sampling + reference-audio cropping)

tsondo:
## Problem

Two runs with the same seed and identical settings produce entirely different songs. The generation seed was only applied to the DiT diffusion stage; two other stochastic stages sampled unseeded randomness:

1. **5Hz LM planning/CoT sampling** — different CoT metadata and audio codes every run whenever `thinking` or CoT is enabled (the default).
2. **Reference-audio segment cropping** — `process_reference_audio()` samples three random 10s windows (front/middle/back) from the reference track with unseeded `random.randint`, so every run conditions on different slices of the reference audio.

### Commit 1: LM sampling

- **`_run_vllm` accepts `seeds` but never uses it.** nano-vllm's sampler draws from the global torch generator (`probs.div_(torch.empty_like(probs).exponential_(1))` in `nanovllm/layers/sampler.py`), which nothing seeds. Now seeds torch/CUDA RNG before `llm.generate`. nano-vllm samples all prompts in a single stream, so per-item seeding isn't possible — the batch is seeded once with `seeds[0]`, which still makes same-seed runs reproducible.
- **`_run_pt` and `_run_mlx` only seed in batch mode.** The single-mode paths (the `batch_size=1` default) skip seeding entirely. Now both single paths seed with `seeds[0]`, mirroring each backend's existing batch-path pattern.
- **`generate_from_formatted_prompt` had no seeds plumbing at all**, and both Phase 1 (CoT) and single-mode Phase 2 (codes) go through it — so those phases were unseeded on every backend. Seeds are now passed via the `cfg` dict and forwarded to all three backends.

### Commit 2: reference-audio cropping

The generation seed (`actual_seed_list[0]`) is threaded into `process_reference_audio()` via a local `random.Random` instance, so same-seed runs condition on the same reference slices. The global RNG is never touched.

## Verification

Tested end-to-end against a running API server (Linux, single RTX 5090, nano-vllm backend): two `/release_task` generations with the same seed — including a reference audio file and CoT metadata generation — now produce **bit-identical audio** (same SHA-256); a different seed produces different audio.

Behavior is unchanged when no seeds are provided (all new code is behind `if seeds:` / `seed is not None`).

Caveat: bit-exact reproducibility still assumes the usual determinism constraints (same hardware, same batch size); these changes make the sampling RNG deterministic, which is the dominant source of run-to-run variation.

## Summary by CodeRabbit

* **New Features**
  * Added consistent seed handling across supported generation backends and audio-generation stages.
  * Repeated generations with the same seed now produce deterministic reference-audio segment selection.
  * Seeded behavior is supported for CoT, audio-code, and reference-audio processing.

* **Tests**
  * Added coverage verifying identical results for matching seeds and different results for different seeds.

### DIFF
--- a/acestep/core/generation/handler/generate_music.py
+++ b/acestep/core/generation/handler/generate_music.py
@@ -331,6 +331,7 @@ class GenerateMusicMixin:
                 actual_batch_size=actual_batch_size,
                 task_type=task_type,
                 flow_edit_morph=flow_edit_morph,
+                seed=actual_seed_list[0] if actual_seed_list else None,
             )
             if audio_error is not None:
                 return audio_error
--- a/acestep/core/generation/handler/generate_music_request.py
+++ b/acestep/core/generation/handler/generate_music_request.py
@@ -106,11 +106,12 @@ class GenerateMusicRequestMixin:
         actual_batch_size: int,
         task_type: str,
         flow_edit_morph: bool = False,
+        seed: Optional[int] = None,
     ) -> Tuple[Optional[List[List[torch.Tensor]]], Optional[torch.Tensor], Optional[Dict[str, Any]]]:
         """Prepare reference/source audio tensors and return early error payload when invalid."""
         if reference_audio is not None:
             logger.info("[generate_music] Processing reference audio...")
-            processed_ref_audio = self.process_reference_audio(reference_audio)
+            processed_ref_audio = self.process_reference_audio(reference_audio, seed=seed)
             if processed_ref_audio is None:
                 return None, None, {
                     "audios": [],
--- a/acestep/core/generation/handler/generate_music_request.py
+++ b/acestep/core/generation/handler/generate_music_request.py
@@ -108,7 +108,11 @@ class GenerateMusicRequestMixin:
         flow_edit_morph: bool = False,
         seed: Optional[int] = None,
     ) -> Tuple[Optional[List[List[torch.Tensor]]], Optional[torch.Tensor], Optional[Dict[str, Any]]]:
-        """Prepare reference/source audio tensors and return early error payload when invalid."""
+        """Prepare reference/source audio tensors and return early error payload when invalid.
+
+        ``seed`` makes the reference-audio segment sampling deterministic so
+        same-seed generations condition on the same reference slices.
+        """
         if reference_audio is not None:
             logger.info("[generate_music] Processing reference audio...")
             processed_ref_audio = self.process_reference_audio(reference_audio, seed=seed)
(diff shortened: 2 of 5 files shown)


tsondo:
Addressed the review findings in the latest commit:

- **MLX hybrid fallback / `torch.multinomial` (llm_inference.py:4098)** — valid catch, fixed. Both the MLX single path and the sequential batch path now seed torch alongside `mx.random`, since the hybrid fallback samples via `_sample_tokens` → `torch.multinomial`.
- **Stale `seeds` docstring (llm_inference.py:1417)** — fixed. The docstring now documents the per-backend contract instead of "TODO: not used yet".
- **Per-item seed semantics in the vllm path (llm_inference.py:918)** — partially addressed. nano-vllm samples all prompts in one shared stream, so true per-item seeding would require sequential per-item generation and forfeit batching throughput. The batch is seeded once with `seeds[0]` (items within a batch still differ from each other; the batch as a whole reproduces), the docstring states this contract explicitly, and `_run_vllm` now logs when distinct per-item seeds are provided so the degradation is visible rather than silent. Happy to switch to per-item sequential generation if maintainers prefer correctness over throughput here.
- **Regression test for seeded segment selection (io_audio.py)** — added `io_audio_seed_test.py`: same seed → identical tensors, different seeds → different tensors.
- **`_prepare_reference_and_source_audio` seed documentation** — added.

Skipped as out of scope for this fix: splitting `llm_inference.py` / `generate_music.py` to meet the 200-line module cap (both files were far over the cap before this PR; this change adds ~40 lines), and the `_prepare_reference_and_source_audio` refactor into separate reference/source helpers.

ChuxiJ (changes requested):
Thanks for closing the main reproducibility gaps. I can't merge this yet because the focused generation-handler suite fails on the current PR head.\n\nLocal verification run:\n\n\n\nFailure:\n\n\n\nThe regression is in :  now calls , but the existing test double still exposes a one-argument lambda. Please update the test double/coverage so the new seed-forwarding contract is exercised by the existing request-preparation tests as well. Once that focused suite is green, the PR should be straightforward to re-review.

ChuxiJ:
Clarifying the failed local verification from my review above, since the code block was mangled by my shell quoting.

Focused run on PR head `abe0600`:

```bash
uv run --no-sync python -m unittest \
  acestep.core.generation.handler.io_audio_seed_test \
  acestep.core.generation.handler.generate_music_request_test \
  acestep.core.generation.handler.generate_music_test
```

Result: 22 tests ran, 1 error.

```text
TypeError: ...<lambda>() got an unexpected keyword argument 'seed'
```

The failing test is `GenerateMusicRequestMixinTests.test_prepare_reference_and_source_audio_returns_error_for_invalid_reference`. The production call changed to `self.process_reference_audio(reference_audio, seed=seed)`, but the existing test double still exposes a one-argument lambda. Please update that test double, and ideally assert that the seed is forwarded through `_prepare_reference_and_source_audio`.

PR #1284, opened 2026-08-02
feat: opt-in on-demand model loading instead of silent fallback to primary

tsondo:
### Summary

Since model selection was introduced (ea865ea), a generation requesting a model that isn't preloaded silently falls back to the primary: the job succeeds, but on different weights than requested, with only a server-console log line. Clients exposing the documented model names as a picker (without `ACESTEP_CONFIG_PATH2/3` configured) appear to work while always generating with the startup model — and preloading every variant costs ~5 GB VRAM each.

This PR adds **opt-in** on-demand loading: with `ACESTEP_ON_DEMAND_MODEL_LOAD=true`, a requested-but-unloaded model loads at request time, replacing the primary in place and downloading the checkpoint on first use. One DiT model in VRAM at a time.

### Scope

Files changed:
- `acestep/api/job_model_selection.py` — env-gated on-demand branch: validate name (`^acestep-v15-[A-Za-z0-9_-]+$`, blocks traversal before any filesystem/network access) → download (handler untouched; failures fall back safely and are cached per name so repeats fail fast) → in-place `initialize_service` swap (failures fail the job — the handler may be torn, so no fallback — and clear the config path so the next request re-attempts a full load). Requesting the already-primary model short-circuits without touching the handler.
- `acestep/api/startup_model_init.py` — capture the primary `initialize_service` kwargs, checkpoint dir, and downloader on `app.state` (as `_service_init_kwargs`, named to avoid colliding with the lazy-init `_model_init_kwargs` shape).
- `acestep/api/job_execution_runtime.py` — handler selection moved inside the executor callable so a multi-second load (or multi-minute first-use download) never blocks the event loop. Side fix: a selection exception previously escaped the try-block and left the job stuck in `running`; it now marks the job failed.
- `acestep/api/job_model_selection_test.py`, `acestep/api/job_execution_runtime_test.py` — targeted tests (see Regression checks).

Explicitly **out of scope**: preloaded multi-model routing (`ACESTEP_CONFIG_PATH2/3` — unchanged, still routes with zero reload), the `/models` init route, Gradio UI, CLI, any refactoring of the selection module.

### Risk and Compatibility

- **Flag off (default): behavior unchanged**, with two deliberate exceptions in the already-degenerate cases: requesting the model that *is* the primary no longer logs a misleading `Model 'X' not found in ['X']` line, and (if a secondary was configured with the same name as the primary) such requests now route to the primary — same model name either way.
- **Non-target platforms unchanged.** The change is hardware-agnostic: no CUDA/MPS/XPU/CPU-specific code; device/dtype/offload decisions remain entirely inside `initialize_service`, and the captured kwargs pass through the same values startup used.
- **Concurrency:** the gate requires both `ACESTEP_QUEUE_WORKERS` and `ACESTEP_API_WORKERS` to resolve to 1 (using the runtime's own `max(1, …)` normalization; unparsable values disable the feature). Rationale: generation jobs serialize on the shared executor, and the `/models` init route can run `initialize_service` from another API executor thread without a lock an executor thread could share. With either >1 the pre-existing fallback behavior applies untouched.
- **Failure semantics:** failures *before* any handler mutation (bad name, cached failure, missing kwargs, download error) fall back to the primary exactly as today. Failures *during* the in-place swap fail the job explicitly rather than generating on a possibly-torn handler — chosen deliberately over silent fallback, since ending silent wrong-model output is this feature's purpose.

### Regression Checks

Automated — `job_model_selection_test.py` (14), `job_execution_runtime_test.py` (3), `job_generation_setup_test.py` (7): **24 passed**. New coverage: gate off preserves fallback; successful load updates handler/config; already-primary short-circuit; invalid-name rejection without download; swap failure fails the job and clears config path; download failure falls back safely and fails fast on repeat; multi-queue-worker and multi-API-worker gates; `"0"` worker normalization; selection exception inside the executor marks the job failed via the fallback handler.

Manual, live single-GPU server (RTX 5090, Linux/CUDA), started with only `acestep-v15-turbo`:
- Requesting `acestep-v15-sft` downloaded (first use)/loaded it and generated on it; `/v1/models` tracked the switch both directions.
- Second sft request: no reload (no on-demand log line, shorter wall time).
- Same seed + parameters across the two models produce different SHA-256 audio — the weights genuinely switch.
- Flag off: requests for unloaded models log the existing fallback line, unchanged.

### Reviewer Notes

Pre-submission, an independent agent ran a commit-scoped adversarial review per CONTRIBUTING.md; dispositions:
- *Torn handler after failed in-place reload* — **Accepted**; swap failures now fail the job and clear the config path (no fallback onto unknown state).
- *Race with `/models` init route under `ACESTEP_API_WORKERS>1`* — **Accepted**; gate extended.
- *Silent wrong-model success on load failure* — **Accepted**; job fails explicitly.
- *`_model_init_kwargs` schema collision* — **Accepted**; renamed to `_service_init_kwargs`.
- *Worker-env parsing edge cases* — **Accepted**; normalization mirrors the runtime, unparsable disables.
- *Repeated stalls on well-formed-but-nonexistent names* — **Accepted**; per-name failure cache.
- *Log-ordering shift for selection messages (now emitted from the executor thread)* — **Accepted, no code change**; cosmetic.
- *Duplicate-name secondary routing change with flag off* — **Rebutted**: same model name resolves either way; configuring a secondary identical to the primary is degenerate. Called out under Risk above.

Known pre-existing issue not addressed (out of scope): the `/models` init route mutates the primary handler under an asyncio lock that generation executor threads do not share; the gate avoids the interaction rather than re-architecting locking.

## Summary by CodeRabbit

* **New Features**
  * Added on-demand model loading for eligible requests, including model validation, cached downloads, and automatic fallback to the primary model.
  * Added retry cooldowns to avoid repeatedly attempting unavailable model downloads.

* **Bug Fixes**
  * Improved job failure reporting and cleanup when model selection or loading fails.

* **Tests**
  * Added coverage for model switching, fallback behavior, validation, retries, worker limits, and disabled loading.


ChuxiJ (approved):
Reviewed the current head after the follow-up fixes. The feature is still opt-in by default, handler selection now runs inside the executor, download failures fall back safely with traceback context and cooldown, and initialize_service failures fail the affected job instead of silently continuing with uncertain handler state.

Focused verification passed locally: job_execution_runtime_test, job_model_selection_test, job_model_selection_on_demand_test, and startup_model_init_test (23 tests).

Issue #1286, opened 2026-08-06
complete/lego tasks ignore src_audio — LM-generated codes silently pre-empt real source-audio conditioning


gurubusvoyage:
## Summary

For `task_type in {"complete", "lego"}`, the LM (5Hz Chain-of-Thought stage) generates
audio codes from text conditioning alone (caption/lyrics/instruction) with **zero
awareness of the provided `src_audio`**. Those LM-generated codes are then passed
downstream as `audio_code_string`, and `_prepare_reference_and_source_audio()`
(`acestep/core/generation/handler/generate_music_request.py`) sees non-empty codes and
takes the "codes already provided" branch — **skipping real source-audio processing
entirely**. The DiT never sees the actual source audio; the result is generation that
bears no audible relationship to the input track, even though `src_audio` was supplied
and the tasks are documented/designed as source-conditioned (`SRC_AUDIO_TASKS`,
`_src_audio_required_tasks`).

## Reproduction

1. Call `generate_music()` with `task_type="complete"`, a valid `src_audio` path, and
   `thinking=True` (the default in most callers).
2. Observe the log line:
   `[generate_music] Audio codes provided, ignoring src_audio and using codes instead`
3. Compare against `task_type="cover-nofsq"` under the same conditions — that task
   correctly logs `Skipping LM for task_type='cover-nofsq' - using DiT directly` followed
   by `[generate_music] Processing source audio...`, and the output is audibly related to
   the source (confirmed at multiple `audio_cover_strength` values, 0.2–0.8).

## Root cause (corrected 2026-08-06)

**Correction**: my original root-cause framing below was wrong. See the "Correction after
reviewing #963/#117" comment for the full explanation — leaving the original text
struck through for context, replaced by the corrected analysis.

LM running for `complete`/`lego` is **intentional design**, not an oversight. PR #963
(merged 2026-03-30, "docs: document LM skip for cover/repaint/extract tasks") documents
this explicitly: *"Think and LM-related CoT parameters apply only to Text-to-Music, Lego,
and Complete tasks"* — i.e. `cover`/`repaint`/`extract` skip the LM by design, but
`complete`/`lego` are meant to use it.

The actual gap is upstream of `skip_lm_tasks`: there is no pipeline that feeds
`src_audio` into the LM before it generates audio codes for `complete`/`lego`. The LM
plans/generates codes from text conditioning alone, blind to the source audio, and
[...]

Issue #117, opened 2026-02-04
Lego task: generated tracks not BPM-locked to source audio grid (timing drift)


andreiostacie:
**Describe the bug**

Lego task generates instrument tracks that are not locked to the source audio's BPM grid. The generated tracks have correct timbre/style but the rhythm drifts and onsets don't align with the beat grid of the source audio. BPM is passed as metadata (`bpm: 135`) but the model treats it as soft text conditioning rather than a hard timing constraint. The generated audio has micro-timing variations that make it unusable for layering over the source beat without manual editing.

**To Reproduce**

1. Load base model (`acestep-v15-base`) with lego task
2. Provide a rhythmically tight source loop (8 bars, 135 BPM, C major, 14.222 seconds)
3. Generate an instrument track (e.g., synth, guitar, piano) using lego task:

```python
params = GenerationParams(
    task_type="lego",
    src_audio="loop2-original.wav",  # tight 135 BPM beat
    caption="electric guitar riff, funk guitar, C major, 135 BPM, house music, instrumental, 4 bars",
    instruction="Generate the GUITAR track based on the audio context:",
    bpm=135,
    keyscale="C major",
    timesignature="4",
    duration=7.111,  # exactly 4 bars at 135 BPM
    inference_steps=80,
    guidance_scale=12.0,
    shift=3.0,
    audio_cover_strength=1.0,
    thinking=False,  # thinking=True crashes on MPS, see additional context
)
```

4. Layer the generated track over the source audio
5. Observe that the generated instrument's rhythmic hits are not aligned to the beat grid — onsets are off by 20-80ms, and the timing drifts across the loop

**Expected behavior**

When providing a rhythmically tight source audio at a specific BPM (135) and explicitly setting `bpm=135` in metadata, the generated lego track should have its rhythmic onsets locked to the same beat grid as the source audio. The track should be usable for direct layering without time-stretching or manual alignment.

For context: Suno's API generates extensions that lock tightly to the source BPM — the timing grid is respected as a hard constraint. ACE-Step's lego task should aim for comparable BPM adherence, especially since the source audio's latents are directly available to the model.

**Attempted workarounds**

1. **Global offset alignment** (cross-correlation of onset envelopes): Shifts the entire generated track to best-fit the beat grid. Helps with the starting offset but doesn't fix internal timing drift.

[...]

PR #1287, opened 2026-08-07
Promote the complete/lego skip-LM workaround to a default (interim)

gurubusvoyage:
## What this changes

Adds `complete` and `lego` to `skip_lm_tasks`, and extends the params-based conditioning
fallback to the same two task types.

```diff
- skip_lm_tasks = {"cover", "cover-nofsq", "repaint", "extract"}
+ skip_lm_tasks = {"cover", "cover-nofsq", "repaint", "extract", "complete", "lego"}
```

```diff
- if params.task_type in ("repaint", "cover", "cover-nofsq", "extract"):
+ if params.task_type in ("repaint", "cover", "cover-nofsq", "extract", "complete", "lego"):
```

## This is a workaround promoted to a default, not a root-cause fix

I want to be explicit about this, because I got the framing wrong the first time and
corrected it publicly in #1286.

Running the LM for `complete`/`lego` is **intentional** — #963 documents that Think/CoT
apply to these tasks. The actual defect is upstream of the skip decision: `src_audio`
never reaches the LM before it generates audio codes, so the LM plans blind to the
source, and those codes then silently pre-empt real source-audio processing in
`_prepare_reference_and_source_audio()` (the *"Audio codes provided, ignoring src_audio"*
branch). @andreiostacie flagged the same mechanism for `lego` in #117 back in February.

The proper fix is to wire `convert_src_audio_to_codes()` / `understand_audio_from_codes()`
— both already exist — into the LM's context for these tasks. **This PR does not do
that.** It changes the default so the source-conditioned behaviour is what users get
until that plumbing lands.

### What this costs

Skipping the LM for these tasks gives up LM CoT caption/metadata enrichment — the same
trade-off #117 describes for `thinking=False`. For source-conditioned tasks I'd argue
that's the right default (following the source matters more than an LM-written caption),
but it is a real loss and should be a deliberate choice, not a silent one.

### Why a per-task gate rather than telling users to set `thinking=False`

`thinking=False` alone does not actually skip the LM:

```python
use_lm = (thinking or need_lm_for_cot) and ... and not skip_lm
need_lm_for_cot = use_cot_caption or use_cot_language or use_cot_metas
```

All three CoT flags default to `True`, so a caller must set **four** flags correctly to
avoid the LM. The Gradio UI's Think checkbox also defaults on. A per-task-type gate is
the only form of this that can't be half-applied.

## Precedent

#886 (*fix: skip LLM for extract tasks*, merged 2026-03-20) made exactly this change for
`extract`, for the same reason — LM-generated conditioning conflicting with a
source-conditioned task. This PR extends that precedent to the two remaining
source-conditioned task types.

## Verification

**Log path.** Before: `complete` logged *"Audio codes provided, ignoring src_audio"* and
never entered source processing. After: it logs the same path as `cover-nofsq`, which
already skipped the LM — `Skipping LM for task_type=...` followed by
`Processing source audio...`.

**Behaviour.** Log parity only shows the plumbing is connected, so I also measured whether
output actually follows the source.

<details>
<summary>Method and numbers</summary>

Method (source-agnostic, reproducible with any source you have a chord chart for): take a
source recording whose chord progression is known independently, run `complete` with
drums+bass as the added tracks, and measure the fraction of frames where the root of the
newly generated bass matches the chart root at the same time position. Two references are
needed to read the number: a **floor**, measured on the source alone with no generated
bass, which captures how much the metric picks up from the source's own low end; and a
**control**, the same run with the bass stem not rendered (−33.5 dB), which should land on
the floor if the metric is measuring the generated track rather than leakage.

| Condition | Root match |
|---|---|
| `complete`, source containing harmony | **56.9%** |
| Control — same run, bass not rendered | 31.5% |
| Floor — source bleed only, no generated bass | 32.5% |

The control landing on the floor is the part that matters: it means the 56.9% comes from
the generated track following the source, not from the measurement seeing the source
through the mix.

</details>

**Scope of verification.** Both task types are log-verified: with the patch applied,
`lego` logs `Skipping LM for task_type='lego' - using DiT directly` followed by
`[generate_music] Processing source audio...`, and never logs the
*"Audio codes provided, ignoring src_audio"* branch. Behaviour (output following the
source harmony) was measured for `complete` only; `lego` shares the identical code path
and #117 flagged the same mechanism for it, but was not separately behaviour-tested.

## Refs

Refs #1286, #117, #963, #886 — interim fix; the fuller fix (feeding src-derived codes
into the LM stage) remains open. Deliberately not `Closes`, since the underlying gap
described in #1286 is not addressed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved music generation for direct-conditioning tasks, including cover, cover-nofsq, repaint, extract, complete, and lego.
  * Preserved provided captions and lyrics during these workflows.
  * Prevented unnecessary language-model processing from overriding source-audio conditioning.
  * Ensured source audio and requested parameters are consistently applied during generation.

### DIFF
--- a/acestep/direct_conditioning_tasks_test.py
+++ b/acestep/direct_conditioning_tasks_test.py
@@ -37,12 +37,24 @@ class RecordingDitHandler:
     """
 
     def __init__(self):
+        """Initialize with no captured call; ``generate_kwargs`` stays None until called."""
         self.generate_kwargs = None
         self.lora_loaded = False
         self.use_lora = False
         self.lora_scale = 1.0
 
     def prepare_seeds(self, batch_size, seed, use_random_seed):
+        """Return a fixed seed list so the run is deterministic.
+
+        Args:
+            batch_size: Number of seeds the caller expects.
+            seed: Ignored; the real handler parses a comma-separated string here.
+            use_random_seed: Ignored; seeds are fixed regardless.
+
+        Returns:
+            ``(seeds, None)`` — a list of ``batch_size`` identical seeds, and the
+            padding info the real handler returns and ``generate_music`` discards.
+        """
         return [1234] * batch_size, None
 
     def generate_music(
@@ -54,6 +66,23 @@ class RecordingDitHandler:
         audio_code_string=None,
         **kwargs,
     ):
+        """Record the conditioning kwargs and stop the pipeline.
+
+        Args:
+            captions: Caption reaching the DiT — ``params.caption`` on a
+                direct-conditioning task, LM-generated otherwise.
+            lyrics: Lyrics reaching the DiT, same origin as ``captions``.
+            src_audio: Source audio path; the value under test.
+            task_type: Task the orchestrator resolved.
+            audio_code_string: LM-generated codes. Non-empty here means codes
+                pre-empted ``src_audio`` downstream.
+            **kwargs: Remaining DiT arguments, captured but not asserted on.
+
+        Returns:
+            A failure result (``success=False``), which makes ``generate_music``
+            return early — so no model runs and nothing is written to disk.
+            The captured kwargs are left on ``self.generate_kwargs``.
+        """
         self.generate_kwargs = {
             "captions": captions,
             "lyrics": lyrics,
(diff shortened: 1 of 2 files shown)


gurubusvoyage:
Declining in this PR — extracting task-routing policy is a broader refactor than this 9-line interim fix warrants, and would swamp the actual change in diff. Happy to do it as a follow-up PR if maintainers want it.

ChuxiJ (approved):
Reviewed this as an interim source-conditioning fix and ran the direct_conditioning_tasks regression tests locally in a clean temporary worktree. The shared direct-conditioning task set keeps LM bypass and params conditioning aligned, including complete and lego. Looks good to merge as the narrow DiT-first workaround.

PR #1290, opened 2026-08-11
fix(repaint): preserve auto chunk mask sentinel

### LINKED ISSUE #1289: bug: chunk_mask_mode="auto" is coerced to a boolean mask
**Did you check DeepWiki?**
- [x] No, this is a source-level tensor dtype bug not covered in the current docs

**Describe the bug**

PR #815 introduced `chunk_mask_mode` with the following contract:

- `explicit`: send the 0/1 chunk mask derived from the repaint interval.
- `auto`: fill the chunk mask with `2.0` so the model decides the chunk boundaries.

However, `_build_chunk_masks_and_src_latents()` creates every per-item mask with `dtype=torch.bool` and stacks them without converting the result to a floating dtype. The later assignment:

```python
chunk_masks_tensor[i] = 2.0
```

therefore coerces `2.0` to boolean `True`. The model receives `1` after its later dtype conversion, not the documented `2.0` auto-mode sentinel. Consequently, `chunk_mask_mode="auto"` does not implement its intended model-decided masking behavior.

This affects every caller that uses the default `auto`, including Gradio Repaint and direct Python/HTTP API generation. The separate boolean `repaint_mask` used for source-latent injection, boundary blending, and waveform splicing is independent and should remain unchanged.

**To reproduce**

```python
chunk_masks, *_ = host._build_chunk_masks_and_src_latents(
    batch_size=1,
    max_latent_length=100,
    repainting_start=[1.0],
    repainting_end=[2.0],
    chunk_mask_modes=["auto"],
    # remaining tensor arguments omitted here
)

assert chunk_masks.dtype == torch.bool
assert torch.all(chunk_masks == 1)
assert not torch.any(chunk_masks == 2)
```

**Expected behavior**

- `auto` returns a floating chunk-mask tensor filled with `2.0`.
- `explicit` returns the repaint interval as floating 0/1 values.
- The independent boolean `repaint_mask`, source-latent handling, and task defaults remain unchanged.

**Proposed minimal fix**

- Convert the stacked model-facing chunk-mask tensor to `float32` before applying the `2.0` sentinel.
- Add focused unit tests for both `auto` (all 2.0) and `explicit` (interval-derived 0/1).
- Document `chunk_mask_mode` and show `explicit` in Repaint inference/API examples so callers can deliberately choose either behavior.

**Scope / compatibility**

The change is limited to the model-facing chunk-mask dtype/value plus related tests and documentation. It does not change the public default (`auto`), request interfaces, the independent repaint enforcement mask, or hardware-specific code paths.

EasonX-30:
## Summary

Fixes #1289.

`chunk_mask_mode="auto"` is documented and implemented as a `2.0` model-decided sentinel. The chunk mask was built as `torch.bool`, however, so assigning `2.0` coerced it to `True`; the model ultimately received `1.0` instead of `2.0`.

This PR converts only the model-facing stacked chunk mask to floating point before applying the sentinel. It also adds focused coverage for both modes and documents the parameter in the English and Chinese inference/API guides.

## Root cause

- Interval masks were created as boolean tensors.
- `chunk_masks_tensor[i] = 2.0` therefore stored `True`, not `2.0`.
- The later model-dtype conversion turned that value into `1.0`, making `auto` behavior differ from the contract introduced in #815.

## Scope

Changed:

- `conditioning_masks.py`: convert the model-facing chunk mask to `float32` before applying auto-mode values.
- `conditioning_masks_test.py`: verify `auto -> 2.0`, `explicit -> interval-derived 0/1`, and the independent `repaint_mask` remains boolean.
- English and Chinese `INFERENCE.md` / `API.md`: document both modes and show explicit-mask Repaint examples.

Out of scope:

- The public `chunk_mask_mode="auto"` default is unchanged.
- Gradio continues to use that default; this PR repairs its intended sentinel behavior rather than changing UI policy.
- Source-latent injection, latent boundary blending, waveform splicing, model code, and hardware-specific paths are unchanged.

## Risk and compatibility

The intentional behavior change is limited to `auto`: it now sends the documented `2.0` sentinel instead of the accidental `1.0`. Explicit masks carry the same 0/1 values as before, now in the floating type expected by model context concatenation. The independent repaint enforcement mask remains `torch.bool`.

Non-target CPU/CUDA/MPS/XPU/MLX branches are unchanged.

## Regression checks

Passed in the requested ACE-Step environment:

```bash
source /home/node59_tmpdata3/yxxia/miniconda3/bin/activate ace-step
python -m unittest \
  acestep.core.generation.handler.conditioning_masks_test \
  acestep.core.generation.handler.conditioning_batch_test \
  acestep.core.generation.handler.conditioning_embed_test
# Ran 20 tests: OK

python -m compileall -q \
  acestep/core/generation/handler/conditioning_masks.py \
  acestep/core/generation/handler/conditioning_masks_test.py

git diff --check
```

A broader handler discovery run executed 298 tests and reported 3 errors plus 1 skip in untouched, unrelated tests (`init_service_test`, `reference_audio_validation_test`, and `vae_decode_chunks_test`). The focused mask/batch/embed suite above is green.

No checkpoint was needed: the regression is deterministic tensor dtype/value handling and is covered without model loading.

### DIFF
--- a/acestep/core/generation/handler/conditioning_masks.py
+++ b/acestep/core/generation/handler/conditioning_masks.py
@@ -35,6 +35,8 @@ class ConditioningMaskMixin:
 
         Returns:
             Tuple of (chunk_masks, spans, is_covers, src_latents, repaint_mask).
+            ``chunk_masks`` is floating point: explicit masks contain 0/1,
+            while auto-mode masks contain the 2.0 model-decided sentinel.
             ``repaint_mask`` is a boolean ``[B, T]`` tensor (True = generate,
             False = preserve source) when any item uses repainting, else ``None``.
         """
@@ -70,7 +72,7 @@ class ConditioningMaskMixin:
             is_cover = (task_type == "cover") or has_code_hint
             is_covers.append(is_cover)
 
-        chunk_masks_tensor = torch.stack(chunk_masks)
+        chunk_masks_tensor = torch.stack(chunk_masks).to(torch.float32)
         if chunk_mask_modes:
             for i, mode in enumerate(chunk_mask_modes):
                 if mode == "auto":
--- a/acestep/core/generation/handler/conditioning_masks_test.py
+++ b/acestep/core/generation/handler/conditioning_masks_test.py
@@ -32,6 +32,7 @@ def _build(
     target_latents: Optional[torch.Tensor] = None,
     repainting_start: Optional[List[float]] = None,
     repainting_end: Optional[List[float]] = None,
+    chunk_mask_modes: Optional[List[str]] = None,
     task_type: str = "",
 ):
     """Call _build_chunk_masks_and_src_latents with sensible defaults."""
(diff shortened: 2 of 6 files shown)


EasonX-30:
Closing this draft because it addressed the wrong scope. The replacement will focus only on making the Gradio Repaint path pass chunk_mask_mode="explicit"; no auto-sentinel dtype change will be included.

PR #1292, opened 2026-08-11
fix(ui): use explicit chunk mask for Repaint

### LINKED ISSUE #1291: bug(ui): Repaint does not pass explicit chunk mask mode
**Did you check DeepWiki?**
- [x] No, this is a source-level UI parameter-routing bug not covered in the current docs

**Describe the bug**

Gradio Repaint asks the user for an exact `repainting_start` and `repainting_end`, but `generate_with_progress()` constructs `GenerationParams` without setting `chunk_mask_mode`.

The omitted field therefore uses the global default:

```python
chunk_mask_mode = "auto"
```

That means selecting Repaint in the UI does not pass the selected interval as the explicit 0/1 model chunk mask. For a UI mode defined by a user-selected edit interval, the request should set `chunk_mask_mode="explicit"`.

The global `auto` default is still useful for other tasks and direct Python/API callers, so it should not be changed.

**To reproduce**

1. Start the Gradio UI.
2. Select **Repaint**.
3. Set a repaint start/end interval and generate.
4. Inspect the `GenerationParams` construction in `acestep/ui/gradio/events/results/generation_progress.py`.
5. Observe that `chunk_mask_mode` is omitted and resolves to `auto`.

**Expected behavior**

- Gradio `task_type="repaint"` passes `chunk_mask_mode="explicit"`.
- All other Gradio task types continue to pass/use `auto`.
- Core and HTTP/Python API defaults remain unchanged.

**Proposed minimal fix**

- Resolve the Gradio chunk-mask mode from the backend task type: `explicit` for `repaint`, `auto` otherwise.
- Pass the resolved value into `GenerationParams`.
- Add focused unit tests for Repaint and non-Repaint task types.
- Document `chunk_mask_mode` and use `explicit` in the English and Chinese Repaint inference/API examples.

**Scope / compatibility**

This changes only Gradio request construction plus documentation. It does not modify core mask construction, direct Python/HTTP defaults, independent repaint enforcement, model code, or hardware-specific paths.

EasonX-30:
## Summary

Fixes #1291.

Gradio Repaint collects an exact `repainting_start` / `repainting_end` interval, but its `GenerationParams` construction omitted `chunk_mask_mode` and therefore used the global `"auto"` default.

This PR makes the Gradio Repaint path pass `chunk_mask_mode="explicit"`, so the selected interval is used as the model-facing 0/1 chunk mask. All other Gradio tasks retain `"auto"`.

## Root cause

The UI mapped the selected generation mode to `task_type="repaint"`, but it did not map that task type to the matching chunk-mask control. `GenerationParams` silently fell back to its global default.

## Scope

Changed:

- Add a focused Gradio task-type resolver: `repaint -> explicit`, all other task types -> `auto`.
- Pass the resolved mode into `GenerationParams` in `generate_with_progress()`.
- Add success and non-target regression tests.
- Document `chunk_mask_mode` and use `explicit` in the English and Chinese Repaint inference/API examples.

Out of scope:

- The `GenerationParams` and HTTP API defaults remain `"auto"`.
- Core chunk-mask construction and `auto` sentinel handling are unchanged.
- Repaint source injection, boundary blending, waveform splicing, model code, and hardware-specific paths are unchanged.

## User impact

Selecting Repaint in Gradio now automatically uses the explicit mask derived from the chosen start/end interval. Python and HTTP API callers can continue choosing either mode directly.

## Regression checks

Checks run locally:

```bash
python -m unittest \
  acestep.ui.gradio.events.results.generation_task_type_test \
  acestep.core.generation.handler.conditioning_masks_test
# Ran 16 tests: OK

python -m unittest discover \
  -s acestep/ui/gradio/events/results \
  -p "*_test.py"
# Ran 98 tests: OK

python -m compileall -q \
  acestep/ui/gradio/events/results/generation_task_type.py \
  acestep/ui/gradio/events/results/generation_progress.py

git diff --check
```

No checkpoint was needed because this is deterministic UI parameter routing, covered by focused unit tests.

## Summary by CodeRabbit

* **New Features**
  * Added task-aware chunk masking for generation, using explicit masking for Repaint tasks and automatic masking elsewhere.
  * Added the `chunk_mask_mode` parameter, supporting `auto` and `explicit` modes.

* **Documentation**
  * Updated English and Chinese API and inference guides with parameter details and revised Repaint examples.

### DIFF
--- a/acestep/ui/gradio/events/results/generation_progress.py
+++ b/acestep/ui/gradio/events/results/generation_progress.py
@@ -27,7 +27,10 @@ from acestep.ui.gradio.events.results.generation_info import (
     DEFAULT_RESULTS_DIR,
     _build_generation_info,
 )
-from acestep.ui.gradio.events.results.generation_task_type import resolve_no_fsq_task_type
+from acestep.ui.gradio.events.results.generation_task_type import (
+    resolve_chunk_mask_mode,
+    resolve_no_fsq_task_type,
+)
 from acestep.ui.gradio.events.results.audio_playback_updates import (
     build_audio_slot_update,
 )
@@ -164,6 +167,7 @@ def generate_with_progress(
         timesteps=parsed_timesteps,
         repainting_start=repainting_start,
         repainting_end=repainting_end,
+        chunk_mask_mode=resolve_chunk_mask_mode(task_type),
         audio_cover_strength=audio_cover_strength,
         cover_noise_strength=cover_noise_strength,
         thinking=think_checkbox,
--- a/acestep/ui/gradio/events/results/generation_task_type.py
+++ b/acestep/ui/gradio/events/results/generation_task_type.py
@@ -15,3 +15,18 @@ def resolve_no_fsq_task_type(task_type: str, no_fsq: bool) -> str:
     if task_type == "cover" and no_fsq:
         return "cover-nofsq"
     return task_type
+
+
+def resolve_chunk_mask_mode(task_type: str) -> str:
+    """Resolve model chunk-mask control for a Gradio task.
+
+    Args:
+        task_type: Resolved backend task type from the Gradio mode.
+
+    Returns:
+        ``explicit`` for Repaint so its selected interval reaches model
+        conditioning; otherwise ``auto`` to preserve existing behavior.
+    """
+    if task_type == "repaint":
+        return "explicit"
+    return "auto"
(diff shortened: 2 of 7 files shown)


ChuxiJ (approved):
Reviewed the Gradio Repaint chunk-mask routing and ran the targeted generation_task_type tests locally in a clean temporary worktree. Repaint now resolves to explicit while other task types keep auto, which matches the intended scope. Looks good to merge.

PR #1296, opened 2026-08-15
fix(api): run lazy model init on executor to keep event loop responsive

### LINKED ISSUE #1281: First lazy-init request blocks the entire API server (health checks included) until model loading finishes
**Did you check DeepWiki?**
- [x] No, this is a new bug not covered in the docs

**Describe the bug**
When the REST API server is started with lazy model loading (`ACESTEP_INIT_SERVICE=false`, the default), the *first* request that triggers model initialization blocks the **entire process** — including `GET /health` and `POST /query_result` for unrelated jobs — for the full duration of model loading (DiT model load, LM tokenizer load, constrained-decoding setup, and especially vLLM's `torch.compile` warm-up on first LM use). In our case this exceeded 90+ seconds and made the server appear completely unreachable to a downstream client, even though it was working correctly the whole time and the job eventually succeeded.

Root cause: in `acestep/api/job_runtime_state.py`, `ensure_models_initialized` calls `do_model_initialization(...)` as a plain synchronous call inside `async def`, with no `await`/`run_in_executor`/`asyncio.to_thread`. Since this coroutine runs on the single-threaded asyncio event loop, that call blocks the whole event loop — and therefore every other in-flight request — until it returns.

```python
async def ensure_models_initialized(app_state: Any) -> None:
    ...
    async with app_state._init_lock:
        ...
        do_model_initialization(app=_AppProxy(app_state), **init_kwargs)  # <- blocks the event loop
```

By contrast, the actual generation step in `acestep/api/job_execution_runtime.py` gets this right, using the same `ThreadPoolExecutor` already sitting on `app_state`:

```python
loop = asyncio.get_running_loop()
result = await loop.run_in_executor(executor, _blocking_generate)
```

**To Reproduce**
1. Start `acestep-api` (or the Docker image) with lazy init left at its default (`ACESTEP_INIT_SERVICE=false`).
2. Submit a `POST /release_task` request with `thinking: true` on a fresh process (nothing loaded yet).
3. Immediately poll `GET /health` or `POST /query_result` from a second client while the first request is being processed.
4. Observe: these calls do not return until model initialization (DiT + LM + vLLM compile) finishes — often 60–120+ seconds on an 8GB-tier GPU — instead of returning promptly with a "still loading"/"queued" status.

**Expected behavior**
`GET /health` and `POST /query_result` should stay responsive during model initialization, the same way they already do once a job is running. A client should be able to tell "still initializing" apart from "server unreachable" without an indefinite/unbounded wait on a single HTTP call.

**Proposed fix**
Wrap the `do_model_initialization(...)` call in `ensure_models_initialized` with the same `loop.run_in_executor(executor, ...)` pattern already used for generation. Flagging this as an issue rather than jumping straight to a PR since it touches a code path shared across CUDA/CPU/MPS/XPU — want to check whether there's a reason it's currently synchronous (e.g. a backend/thread-safety constraint specific to model init) before assuming `run_in_executor` is safe everywhere. Happy to submit the PR once that's confirmed.

**Screenshots**
N/A — this is a backend/API timing issue, not a UI issue.

**Desktop (please complete the following information):**
 - OS: Windows 11 (Docker Desktop, WSL2 backend)
 - Deployment: `ghcr.io/ace-step/ace-step-1.5:latest` (Docker), CUDA backend
 - GPU: RTX 4060 Laptop, 8GB (tier3), `acestep-v15-turbo` + `acestep-5Hz-lm-0.6B`

**Smartphone (please complete the following information):**
N/A

**Additional context**
Confirmed via process inspection during the hang that the worker process stayed in a normal running state the whole time (not deadlocked/crashed) — this is a scheduling/blocking issue, not a crash. `torch._inductor` compile-worker subprocesses were actively spinning during the slow window, consistent with vLLM's first-use `torch.compile` cost. Not yet reproduced/checked on CPU/MPS/XPU backends.

glb99:
ensure_models_initialized() called do_model_initialization() synchronously inside an async def while holding the init lock, blocking the single-threaded event loop for the full duration of model loading (DiT load, LM tokenizer, vLLM torch.compile warm-up — 60-120+ s on 8GB-tier GPUs). Every other in-flight request, including GET /health and POST /query_result, hung until init finished, making the server indistinguishable from an unreachable one.

Offload the call to app.state.executor via loop.run_in_executor(), the same pattern job_execution_runtime.py already uses for generation. The _init_lock / double-check / _init_error semantics are unchanged: state flags are set inside do_model_initialization itself, and concurrent callers still wait on the async lock — but the event loop stays free.

Add a deterministic regression test: the fake init blocks in the worker thread on a threading.Event that only the event loop can set, so inline (blocking) execution deadlocks and times out while offloaded execution completes immediately; also asserts init ran off the loop thread.

Fixes #1281

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsiveness during model initialization by preventing blocking work from freezing the event loop.
  * Model setup now runs safely in the background, allowing other asynchronous tasks to continue without interruption.

* **Tests**
  * Added coverage confirming initialization runs outside the event-loop thread and remains non-blocking.

### DIFF
--- a/acestep/api/job_runtime_state.py
+++ b/acestep/api/job_runtime_state.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import os
 from typing import Any, Optional
 
@@ -14,6 +15,10 @@ async def ensure_models_initialized(app_state: Any) -> None:
     If models were already loaded at startup (``ACESTEP_NO_INIT=false``), this
     returns immediately.  Otherwise it performs on-demand initialization using
     the kwargs stored on ``app_state._model_init_kwargs`` during lifespan setup.
+
+    Initialization runs on ``app_state.executor`` rather than inline: it is
+    slow, blocking work that would otherwise freeze the event loop and hang
+    every other in-flight request (e.g. ``/health``) until it finishes.
 
     Args:
         app_state: Application state object containing initialization flags.
@@ -58,7 +63,11 @@ async def ensure_models_initialized(app_state: Any) -> None:
             def __init__(self, state: Any) -> None:
                 self.state = state
 
-        do_model_initialization(app=_AppProxy(app_state), **init_kwargs)
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(
+            app_state.executor,
+            lambda: do_model_initialization(app=_AppProxy(app_state), **init_kwargs),
+        )
 
 
 async def cleanup_job_temp_files(app_state: Any, job_id: str) -> None:
--- a/acestep/api/job_runtime_state_test.py
+++ b/acestep/api/job_runtime_state_test.py
@@ -95,6 +95,7 @@ class JobRuntimeStateTests(unittest.IsolatedAsyncioTestCase):
         init_threads: list[threading.Thread] = []
 
         def _blocking_init(*_args: object, **_kwargs: object) -> None:
+            """Record the calling thread, then block until the event loop sets the event."""
             init_threads.append(threading.current_thread())
             if not loop_alive_during_init.wait(timeout=0.5):
                 raise RuntimeError(
--- a/acestep/api/job_runtime_state_test.py
+++ b/acestep/api/job_runtime_state_test.py
@@ -5,7 +5,9 @@ from __future__ import annotations
 import asyncio
 import os
 import tempfile
+import threading
 import unittest
+from concurrent.futures import ThreadPoolExecutor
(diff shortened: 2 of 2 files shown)


ChuxiJ (approved):
Reviewed the focused lazy-init executor change and ran the targeted job_runtime_state tests locally in a clean temporary worktree. The fix keeps the existing lock/error semantics while moving blocking model initialization off the event loop. Looks good to merge.

PR #1300, opened 2026-08-24
fix(api): failed job cache payload carries the recorded error

FlexOr2:
Stacked on #1091 (VRAM preflight opt-out) and #1092 (DiT param exposure) — this diff includes their commits until those land; the actual new work here starts at `fix(api): failed job cache payload carries the recorded error`.

## Motivation

`/query_result` serves the local cache before the job store, so a failing
job's cause never reached the client: `mark_failed` records the traceback on
the store record, but the cached failure payload only carried `stage` and
`progress`. Callers saw `status: 2` with no error text and had to grep server
logs to learn why a generation failed.

## Change

- The failed branch of `update_local_cache` now includes the store record's
  `error`, mirroring the store-derived payload already produced by
  `query_result_service`.
- Review cosmetics folded in as a follow-up commit: the record was being
  fetched from the store twice per call (once for `env`/`create_time`, once
  for `error`) — now fetched once and threaded through both helpers. The
  `error` key is only added when `status == "failed"`, not in every
  non-succeeded branch, since it's only meaningful there.

## Testing

`python -m unittest acestep.api.jobs.local_cache_updates_test` — 7 tests,
covering the new failed-payload-carries-error case and the pre-existing cases
(succeeded/extract-codes/progress/etc.) unaffected by the refactor.


ChuxiJ (changes requested):
Thanks for the fixes here. I ran the focused tests locally in a clean temporary worktree and they pass:

`local_cache_updates_test`, `release_task_models_test`, `release_task_param_parser_test`, `release_task_request_builder_test`, `job_generation_setup_test`, and `generate_music_test` — 43 tests total.

I still do not want to merge this PR as-is because it combines several unrelated changes under the failed-cache title:

- failed local-cache payload now carries the recorded error;
- `/release_task` starts exposing sampler / velocity / latent controls;
- generation adds CUDA cleanup plus `ACESTEP_SKIP_VRAM_PREFLIGHT` behavior.

Those are each useful, but they have different risk surfaces. Please split this into smaller PRs, ideally starting with the failed-cache error propagation only. For the API parameter PR, please also validate `sampler_mode` as an enum/Literal (`"euler"` / `"heun"`) and add the invalid-value test CodeRabbit suggested. For the cache helper, please avoid eager `now_fn()` evaluation when `created_at` already exists.

Happy to re-review the split cache PR first; that one should be very quick.

PR #1301, opened 2026-08-24
fix(gpu_config): size the LM's DiT reserve for real, duration-aware demand

FlexOr2:
Stacked on #1091 (VRAM preflight opt-out) and #1092 (DiT param exposure) — this diff includes their commits until those land; the actual new work here starts at `fix(gpu_config): size the LM reserve for the DiT's real, duration-aware demand`.

## Motivation

The VRAM pre-flight and the LM allocator measure the same GPU with different
rulers. `gpu_config.py:get_lm_gpu_memory_ratio` kept a *constant*
`dit_reserve_gb = 1.5` free for the DiT, while the pre-flight demands
`DIT_INFERENCE_VRAM_PER_BATCH[dit_key] * batch * duration/60 + 0.5`. The two
agree at 120s and diverge after that — with an LM resident, the DiT was left
1.30 GB where a 165s take needs 1.88 GB, so the pre-flight rejected long
xl-turbo takes that the allocator had already promised room for.

## Change

- **The reserve is duration- and batch-aware.** `get_dit_inference_reserve_gb()`
  mirrors the pre-flight's own formula and adds a fixed headroom margin. The
  LM initializes before any request, so the reserve is sized for the longest
  track and batch size the deployment intends to render with the LM
  resident (a GPU tier's `max_duration_with_lm`), overridable via env vars.
  The resident DiT's checkpoint path now reaches the LM init, so an XL DiT
  reserves XL activations instead of the default profile.
- **The reserve comes out of the KV cache, never out of the LM weights — and
  never below what the LM can legally need.** The floor is what nano-vllm's
  scheduler actually requires: blocks for a full `max_model_len` window,
  doubled because classifier-free guidance schedules the conditional and
  unconditional sequence as a pair. The floor is derived from the
  checkpoint's own `config.json` (layers × KV heads × head dim × dtype), not
  from an estimate table. When the reserve would cut into that floor, the
  shortfall is logged with the track length that *is* supported; when even
  the floor doesn't fit, the LM fails to initialize with `LmKvCacheTooSmallError`
  instead of the DiT hitting an OOM or nano-vllm raising "Insufficient KV
  cache to schedule sequence" mid-generation.
- **Memory the allocator cannot see is subtracted up front.** nano-vllm sizes
  the KV cache against PyTorch's allocator bookkeeping, which misses the CUDA
  context and fragmentation the driver reports. Without that correction every
  gigabyte reserved for the DiT was spent twice.

## Measured (RTX 3090, 23.53 GiB usable, xl-turbo + `acestep-5Hz-lm-4B`, 165s track)

| | before | after |
|---|---|---|
| LM ratio | 0.915 → clamped to 0.900, silently | 0.836, clamp not reached |
| KV cache | 2.46 GB / 17.9k tokens (≈2 CFG pairs) | 1.13 GB / 8.2k tokens (one CFG pair, 4096-token windows) |
| free for the DiT | 1.28 GB | 2.43 GB |
| longest xl-turbo take that passes, batch 1 | ~96s | ~232s |

Real DiT peak measured with the pre-flight bypassed to let the run complete:
23415 MiB of 24576 MiB against a 22813 MiB plateau while the LM ran — **0.59
GB** actual DiT usage where the pre-flight demands 1.88 GB for the same
inputs. The pre-flight stays conservative by roughly a factor of three; this
fix makes the LM respect that conservatism instead of contradicting it.

## Testing

Targeted unit tests added alongside the change (`gpu_config_lm_reserve_test.py`,
`startup_llm_init_test.py`, `llm_backend_compat_test.py`) lock the reserve
formula, the KV floor derivation, and the `LmKvCacheTooSmallError` path.


ChuxiJ (changes requested):
Thanks for tackling the LM/DiT memory budgeting problem. I cannot merge this in its current form for two reasons.

First, the PR is stacked on older `main` (`14c0211`) and brings in several independent changes before the actual LM reserve work:

- `ACESTEP_SKIP_VRAM_PREFLIGHT` / CUDA cleanup work from #1091
- `/release_task` sampler/DiT parameter exposure from #1092
- failed-job cache error payload work from #1300
- then the LM reserve / KV-cache admission changes

Please rebase onto current `main` and split this down so the PR under review contains only the LM reserve / KV-cache admission behavior. The other pieces should land through their own PRs if still desired.

Second, the focused verification is not green on the current PR head `9c2cc9d`.

```bash
uv run --no-sync python -m unittest \
  acestep.gpu_config_lm_reserve_test \
  acestep.api.startup_llm_init_test \
  acestep.llm_backend_compat_test \
  acestep.llm_inference_enforce_eager_test
```

Result: 38 tests ran, 3 failures:

- `LlmInitializeBackendCompatTests.test_initialize_reports_failure_when_the_kv_cache_does_not_fit`
- `TestEnforceEagerWhenFlashAttnMissing.test_enforce_eager_still_true_for_jetson_even_with_flash_attn`
- `TestEnforceEagerWhenFlashAttnMissing.test_enforce_eager_true_when_flash_attn_missing`

The first failure reports a PyTorch fallback unpacking error instead of the expected KV-cache failure, and the enforce-eager tests now see `None` where `True` is expected. Please fix those before re-review.

PR #1302, opened 2026-08-25
fix(inference): keep requested vocal_language over CoT detection

badaldesai:
### Summary

`generate_music` replaced a caller's requested `vocal_language` with the LM's
CoT-detected value whenever `use_cot_language` was set, which is the default
(`release_task_models.py`, `release_task_request_builder.py`). The overwrite was
unconditional, unlike the neighbouring `_update_metadata_from_lm`, which fills
bpm / keyscale / duration only when the caller left them unset.

The language reaches the model as a plain-text header in the lyric branch
(`_format_lyrics` in `core/generation/handler/prompt_utils.py`), so an overridden
language conditions generation under a different phonetic prior. A request for
`pa` whose romanized lyrics the LM read as `hi` generated as Hindi, with nothing
signalling that the request had been overridden.

The LM is not told the requested language either: `user_metadata` carries only
bpm / keyscale / timesignature / duration, so constrained decoding cannot pin the
field even though `pa` is in `VALID_LANGUAGES`.

### Scope

Changed:
- `acestep/inference.py` — new `_is_vocal_language_unset` helper; the CoT
  language assignment is guarded by it. A detected language that disagrees with
  an explicit request is logged rather than applied.
- `acestep/vocal_language_cot_test.py` — new tests.

Explicitly out of scope:
- The `use_cot_caption` branch, which keeps its existing behaviour.
- HTTP request defaults (`vocal_language` stays `"en"`, `use_cot_language` stays
  `True`), so no public interface changes.
- The training/inference lyric-prompt mismatch. Separate issue, noted below.

### Risk and Compatibility

Target path is the LM/CoT metadata merge in `generate_music`. Non-target paths
unchanged: no hardware or runtime code is touched, so CPU / CUDA / ROCm / MPS /
XPU / MLX paths are unaffected. The change is pure control flow over two Python
strings.

Behaviour change is limited to one case: an explicit `vocal_language` is now
preserved instead of overwritten. Callers that send nothing, `""`, or
`"unknown"` keep auto-detection exactly as before — `"unknown"` is the
`GenerationParams.vocal_language` default and is treated as unset.

### Regression Checks

- Full suite: **1211 tests / 7 failures / 24 errors** on `upstream/main`, and
  **1218 / 7 / 24** with this change. The +7 are the new tests; failure and
  error counts are identical. Baseline was re-measured on the same environment
  rather than quoted from an earlier run.
- `python -m unittest discover -s . -p "test_*.py"` — 13 tests, OK.
- Scenarios validated: explicit `pa` / `hi` / `en` / `ur` / `yue` preserved;
  unset, empty and `"unknown"` still auto-detect; `"unknown"` matching is case-
  and whitespace-insensitive; a padded real code (`"  pa  "`) still counts as a
  choice.
- The tests also pin the two defaults the guard depends on
  (`GenerationParams.vocal_language == "unknown"`,
  `GenerateMusicRequest.use_cot_language is True`) so a future default change
  cannot silently alter the guard's meaning.

### Reviewer Notes

Pre-existing, not addressed: the 7 failures / 24 errors above are present on
`upstream/main` and live in the gradio UI, mlx and cfg_scale modules.

Follow-up: training and inference disagree on the lyric prompt. Inference sends
``# Languages\n{lang}\n\n# Lyric\n{lyrics}<|endoftext|>``; training's
`encode_lyrics` (`training/dataset_builder_modules/preprocess_lyrics.py`)
tokenizes bare lyrics with no header and no terminator, at `max_length=512`
against inference's 2048. A grep for `# Languages` returns only inference-side
hits. Per-sample `language` is read into `AudioSample.language` and saved to
`.pt` metadata but never enters a tensor, so language conditioning is not
learned during fine-tuning. Happy to open that as a separate issue — it changes
the meaning of cached preprocessed tensors and would invalidate existing
datasets, so it looks like a maintainer call rather than something to bundle
here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Explicitly selected vocal languages are now preserved.
  * Automatic language detection applies when the language is omitted, blank, or set to “unknown.”
  * Omitted language requests remain unset until detection or fallback handling occurs.
  * Improved reporting when detected and requested languages conflict.
  * Singular seed values are now retained correctly.

* **Tests**
  * Added coverage for language aliases, explicit language codes, placeholder values, and case-insensitive matching.
  * Verified vocal-language behavior across the complete request and generation flow.


ChuxiJ (changes requested):
Thanks for tackling the explicit-language overwrite; preserving a caller-selected language is the right direction.

I do not think this is merge-ready yet because the API path cannot currently distinguish an omitted `vocal_language` from an explicitly supplied `"en"`. `GenerateMusicRequest` defaults `vocal_language` to `"en"`, and this PR makes `_is_vocal_language_unset("en")` false, so a `/release_task` caller that omits `vocal_language` but keeps `use_cot_language=True` will no longer allow CoT language detection. That is a behavior change from the previous API default path and can silently pin non-English prompts to English.

Please carry the explicit-vs-default distinction through request construction/setup, for example by using Pydantic `model_fields_set` or a request-builder marker, and add an integration-style test for both cases:

- omitted `vocal_language` still allows CoT detection;
- explicitly supplied `vocal_language="en"` is preserved over a conflicting CoT detection.

The helper-level tests are useful, but this request-surface distinction needs coverage before merge.

PR #1305, opened 2026-08-27
fix(generation): neutralize cover-only params for non-cover tasks

### LINKED ISSUE #1271: Cover-mode cover_noise_strength silently persists into text2music, producing full-duration static
> **📚 Before submitting:** Have you tried [DeepWiki AI Assistant](https://deepwiki.com/ace-step/ACE-Step-1.5)?

**Did you check DeepWiki?**
- [x] Yes, I checked DeepWiki but couldn't find a solution

**Describe the bug**

After running a Cover/Remix generation in the Gradio UI, the cover-related sliders — specifically `cover_noise_strength` — keep their values when switching back to text2music. With `cover_noise_strength > 0` in text2music, every generation produces a full-duration featureless roar (steady noise, no musical structure) regardless of caption, lyrics, seed, or thinking mode.

This is easy to misdiagnose: in my case the first generations after a Remix session happened to be instrumentals, so it initially looked like "instrumental generation is broken." It isn't — vocal text2music fails identically with these settings.

**To Reproduce**

1. Custom mode, text2music: generate any song → works.
2. Run a Remix/Cover generation (UI sets `audio_cover_strength=0`, `cover_noise_strength=0.2`; sidecar JSONs confirm).
3. Switch back to text2music and generate anything (with or without lyrics, thinking on or off).
4. Output is steady noise for the entire duration. The generation-params JSON saved alongside the output shows `task_type: "text2music"` with `cover_noise_strength: 0.2` still applied.

**Expected behavior**

Cover-only parameters should be reset or ignored when `task_type` is text2music (or the UI should warn). Text2music output should be unaffected by settings from a previous cover session.

**Isolation evidence (controlled matrix)**

Six 60s generations via the Python API (`generate_music`), fixed seed 12345, `acestep-v15-turbo` + `acestep-5Hz-lm-1.7B`, caption "epic orchestral instrumental, war drums, strings, horns". Verdicts computed from the waveform (RMS-envelope coefficient of variation; music ≈ 0.4–0.9, the failure mode ≈ 0.02–0.04):

| Test | lyrics | thinking | audio_cover_strength | cover_noise_strength | Result |
|---|---|---|---|---|---|
| A | `[Instrumental]` | on | 0.0 | 0.2 | **steady noise** |
| B | `[Instrumental]` | on | 1.0 | 0.0 | music |
| C | `[inst]` | on | 1.0 | 0.0 | music |
| D | `[Instrumental]` | off | 1.0 | 0.0 | music |
| E | `[Instrumental]` | on | **0.0** | 0.0 | music |
| F | `[Instrumental]` | on | 1.0 | **0.2** | **steady noise** |

E vs F isolates `cover_noise_strength` as the single causal variable; `audio_cover_strength=0` alone is harmless.

<details>
<summary>Repro script (loads models once, runs the six tests, computes verdicts programmatically)</summary>

```python
"""Run from the repo root: .venv/Scripts/python test_instrumental_matrix.py"""
import os
import numpy as np
import soundfile as sf

PROJECT_ROOT = os.path.dirname(os.path.abspath(__file__))
OUT_DIR = os.path.join(PROJECT_ROOT, "test_matrix_outputs")
os.makedirs(OUT_DIR, exist_ok=True)

CAPTION = "epic orchestral instrumental, war drums, strings, horns"
SEED = 12345
DURATION = 60

# name, lyrics, thinking, audio_cover_strength, cover_noise_strength
TESTS = [
    ("A_baseline_repro",   "[Instrumental]", True,  0.0, 0.2),
    ("B_fix_both",         "[Instrumental]", True,  1.0, 0.0),
    ("C_inst_token",       "[inst]",         True,  1.0, 0.0),
    ("D_lm_off",           "[Instrumental]", False, 1.0, 0.0),
    ("E_only_covstr_0",    "[Instrumental]", True,  0.0, 0.0),
    ("F_only_covnoise_02", "[Instrumental]", True,  1.0, 0.2),
]

def analyze(path):
    x, sr = sf.read(path, always_2d=True)
    x = x.mean(axis=1)
    rms = float(np.sqrt(np.mean(x**2)))
    crest = float(np.max(np.abs(x))) / (rms + 1e-10)
    w = int(sr * 0.5)
    env = np.array([np.sqrt(np.mean(x[i:i + w] ** 2)) for i in range(0, len(x) - w, w)])
    env_cv = float(np.std(env) / (np.mean(env) + 1e-10))
    verdict = "STEADY-NOISE" if (env_cv < 0.25 and crest < 4.0) else "MUSIC"
    return dict(rms=rms, crest=crest, env_cv=env_cv, verdict=verdict)

def main():
    from acestep.handler import AceStepHandler
    from acestep.llm_inference import LLMHandler
    from acestep.inference import GenerationParams, GenerationConfig, generate_music

    dit_handler = AceStepHandler()
    dit_handler.initialize_service(project_root=PROJECT_ROOT,
                                   config_path="acestep-v15-turbo", device="cuda")
    llm_handler = LLMHandler()
    llm_handler.initialize(checkpoint_dir=os.path.join(PROJECT_ROOT, "checkpoints"),
                           lm_model_path="acestep-5Hz-lm-1.7B", backend="vllm", device="cuda")

    for name, lyrics, thinking, cov_str, cov_noise in TESTS:
        params = GenerationParams(caption=CAPTION, lyrics=lyrics, duration=DURATION,
                                  seed=SEED, thinking=thinking,
                                  audio_cover_strength=cov_str,
                                  cover_noise_strength=cov_noise)
        config = GenerationConfig(batch_size=1, use_random_seed=False,
                                  seeds=[SEED], audio_format="wav")
        result = generate_music(dit_handler, llm_handler, params, config, save_dir=OUT_DIR)
        if result.success and result.audios:
            print(name, analyze(result.audios[0]["path"]))
        else:
            print(name, "FAILED:", result.error)

if __name__ == "__main__":
    main()
```

</details>

**Mechanism**

In [`acestep/models/turbo/modeling_acestep_v15_turbo.py` (~L2050)](https://github.com/ace-step/ACE-Step-1.5/blob/6d467e4b5081ccb0abf1ec1bf4fdf9051a2d34b0/acestep/models/turbo/modeling_acestep_v15_turbo.py#L2050), `cover_noise_strength > 0` initializes `xt = renoise(src_latents, nearest_t, noise)` and truncates the timestep schedule to start at `t ≈ 1 − cover_noise_strength`. In cover mode `src_latents` is real audio, so this is correct. In text2music `src_latents` is the **silence latent**, so the sampler starts from noise-blended-silence at t≈0.8 on an already short 8-step turbo schedule and never converges.

**Suggested fix direction**

Force `cover_noise_strength=0` (and other cover-only params) when `task_type == "text2music"` in the generation handler, or reset the sliders on mode switch in the Gradio UI. Possibly related to #988, which reported the reverse direction (mode switch to Remix not updating `task_type`, also yielding noise) and was closed stale — both suggest per-mode parameter hygiene on mode switches.

**Desktop (please complete the following information):**
- OS: Windows 11 Pro
- GPU: RTX 5080 16 GB (torch 2.7.1+cu128, Python 3.12, uv install)
- Models: acestep-v15-turbo + acestep-5Hz-lm-1.7B (LM on PyTorch backend — vllm falls back on Windows/no Triton)
- Repo commit: `6d467e4b5081ccb0abf1ec1bf4fdf9051a2d34b0`
- Browser: Chrome (UI bug is state-persistence, not browser-specific)

**Additional context**

Sidecar JSONs from the failing UI sessions available on request.

PythonToGo:
### Summary

  - Cover-only conditioning params (audio_cover_strength / cover_noise_strength) could leak from a Remix/Cover session
  into text2music via saved or stale UI state, producing full-duration noise regardless of prompt, lyrics, or seed (#1271).
  - Root cause: text2music is the only task type whose source latent is silence rather than real audio content. Blending
  noise toward a silence latent via cover_noise_strength never converges on an 8-step turbo schedule, producing steady
  noise.
  - Fix: reset both params to their neutral defaults (1.0 / 0.0) specifically when task_type == "text2music", in the shared generate_music handler used by every entry point (single/batch/API). Also reset the cover_noise_strength slider's displayed value (not just hide it) when leaving cover mode in the Gradio UI.

  ### Scope

  - Files changed: generate_music.py (call site), generate_music_request.py (_neutralize_cover_only_params helper), generate_music_request_test.py (unit tests), mode_ui.py (slider reset on mode switch).
  - Out of scope: any other task type's conditioning logic. cover, cover-nofsq, repaint, lego, extract, and complete are all left untouched by this guard, since they operate on real source audio/codes where these params are legitimate, documented controls, not cover leftovers.

  ### Risk and Compatibility

  - The guard triggers only for task_type == "text2music"; every other task type is unaffected — confirmed by unit test
  test_neutralize_cover_params_preserved_for_src_audio_tasks, which asserts pass-through for all six non-text2music task
  types.
  - Self-review note: an earlier version of this fix used an allow-list of ("cover", "cover-nofsq") instead of a
  deny-list of "text2music". That would have silently reset audio_cover_strength/cover_noise_strength for
  repaint/lego/extract/complete callers passing those params intentionally (e.g. via direct API use) — a regression
  outside the issue's scope. Caught during self-review before opening this PR and fixed by inverting the condition to target
  only text2music.
  - Non-target platforms/paths unchanged;  this is a pure request-normalization change; no CUDA/CPU/MPS-specific code
  touched.

  ### Regression Checks

  - `python -m unittest acestep.core.generation.handler.generate_music_request_test -v`   —  12/12 pass, including 2 tests covering both the text2music reset and the preserved-params case across all 6 non-text2music task types.
  - Traced modeling_acestep_v15_turbo.py / modeling_acestep_v15_base.py to confirm text2music is the only task type using a silence src_latents, which is the actual root cause of the reported noise.

  ### Reviewer Notes
  - Known minor issue, not addressed here: in mode_ui.py, leaving cover mode resets the cover_noise_strength slider's
  displayed value but not audio_cover_strength's (it's only forced to 0.0 on entering cover mode). This is cosmetic
  only; the backend guard already neutralizes the value at generation time regardless of what the slider displays,  so
  left as a follow-up rather than expanding this PR's scope.

  Fixes #1271

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented cover-specific settings from carrying over into text-to-music generation.
  * Reset cover noise strength to its neutral value when leaving Cover mode.
  * Preserved user-provided cover settings for supported workflows, including Cover, Cover No-FSQ, Repaint, Lego, Extract, and Complete modes.
  * Improved consistency when switching between generation modes, preventing stale settings from affecting new music generation requests.

### DIFF
--- a/acestep/core/generation/handler/generate_music.py
+++ b/acestep/core/generation/handler/generate_music.py
@@ -296,8 +296,9 @@ class GenerateMusicMixin:
             instruction=instruction,
         )
 
-        # Drop stale cover-only params (e.g. leaked from a Remix session) for non-cover tasks;
-        # all entry points (single/batch/API) pass through here (issue #1271).
+        # Neutralize stale cover-only params for text2music; preserve them for
+        # source-audio tasks. All entry points (single/batch/API) pass through
+        # here (issue #1271).
         audio_cover_strength, cover_noise_strength = self._neutralize_cover_only_params(
             task_type=task_type,
             audio_cover_strength=audio_cover_strength,
--- a/acestep/core/generation/handler/generate_music.py
+++ b/acestep/core/generation/handler/generate_music.py
@@ -296,6 +296,14 @@ class GenerateMusicMixin:
             instruction=instruction,
         )
 
+        # Drop stale cover-only params (e.g. leaked from a Remix session) for non-cover tasks;
+        # all entry points (single/batch/API) pass through here (issue #1271).
+        audio_cover_strength, cover_noise_strength = self._neutralize_cover_only_params(
+            task_type=task_type,
+            audio_cover_strength=audio_cover_strength,
+            cover_noise_strength=cover_noise_strength,
+        )
+
         # Turbo models bake guidance into the distillation process and do not
         # use CFG.  Forcing guidance_scale to 1.0 avoids double-application of
         # guidance that produces noise or NaN/Inf on float16 (see issue #927).
--- a/acestep/core/generation/handler/generate_music_request.py
+++ b/acestep/core/generation/handler/generate_music_request.py
@@ -55,6 +55,25 @@ class GenerateMusicRequestMixin:
             return "cover", TASK_INSTRUCTIONS["cover"]
         return task_type, instruction
 
+    def _neutralize_cover_only_params(
+        self,
+        task_type: str,
+        audio_cover_strength: float,
+        cover_noise_strength: float,
+    ) -> Tuple[float, float]:
+        """Reset cover-only params to their neutral defaults for non-cover tasks.
+
+        Stale UI state (e.g. after a Remix session) can leak a non-zero
+        ``cover_noise_strength`` into ``text2music`` and produce full-duration
+        noise (issue #1271). Run after ``_resolve_generate_music_task`` so the
+        audio-codes auto-switch to ``cover`` is preserved.
+        """
+        if task_type in ("cover", "cover-nofsq"):
+            return audio_cover_strength, cover_noise_strength
+        if audio_cover_strength != 1.0 or cover_noise_strength != 0.0:
+            logger.info("[generate_music] Neutralizing cover-only params for task_type={!r}", task_type)
+        return 1.0, 0.0
+
     def _prepare_generate_music_runtime(
         self,
         batch_size: Optional[int],
--- a/acestep/core/generation/handler/generate_music_request.py
+++ b/acestep/core/generation/handler/generate_music_request.py
@@ -61,14 +61,20 @@ class GenerateMusicRequestMixin:
         audio_cover_strength: float,
         cover_noise_strength: float,
     ) -> Tuple[float, float]:
-        """Reset cover-only params to their neutral defaults for non-cover tasks.
+        """Reset cover-only params to their neutral defaults for text2music.
 
         Stale UI state (e.g. after a Remix session) can leak a non-zero
         ``cover_noise_strength`` into ``text2music`` and produce full-duration
-        noise (issue #1271). Run after ``_resolve_generate_music_task`` so the
-        audio-codes auto-switch to ``cover`` is preserved.
+        noise, because ``text2music`` is the only task type whose source
+        latent is silence rather than real audio content (issue #1271).
+        Every other task type (cover, cover-nofsq, repaint, lego, extract,
+        complete) operates on real source audio/codes, where these params
+        are meaningful, documented controls -- not just cover leftovers --
+        so they must be left untouched. Run after
+        ``_resolve_generate_music_task`` so the audio-codes auto-switch to
+        ``cover`` is preserved.
         """
-        if task_type in ("cover", "cover-nofsq"):
+        if task_type != "text2music":
             return audio_cover_strength, cover_noise_strength
         if audio_cover_strength != 1.0 or cover_noise_strength != 0.0:
             logger.info("[generate_music] Neutralizing cover-only params for task_type={!r}", task_type)
--- a/acestep/core/generation/handler/generate_music_request_test.py
+++ b/acestep/core/generation/handler/generate_music_request_test.py
@@ -47,6 +47,28 @@ class GenerateMusicRequestMixinTests(unittest.TestCase):
         self.assertEqual(task, "cover")
         self.assertNotEqual(instruction, "old")
 
+    def test_neutralize_cover_params_reset_for_non_cover_tasks(self):
+        """Non-cover tasks must drop leaked cover params to neutral (issue #1271)."""
+        host = _Host()
+        for task in ("text2music", "repaint", "lego", "extract"):
+            strength, noise = host._neutralize_cover_only_params(
+                task_type=task,
+                audio_cover_strength=0.0,
+                cover_noise_strength=0.2,
+            )
+            self.assertEqual((strength, noise), (1.0, 0.0), task)
+
+    def test_neutralize_cover_params_preserved_for_cover_tasks(self):
+        """Cover tasks must keep their cover conditioning params."""
+        host = _Host()
+        for task in ("cover", "cover-nofsq"):
+            strength, noise = host._neutralize_cover_only_params(
+                task_type=task,
+                audio_cover_strength=0.3,
+                cover_noise_strength=0.15,
+            )
+            self.assertEqual((strength, noise), (0.3, 0.15), task)
+
     def test_prepare_runtime_normalizes_batch_and_duration(self):
         """Runtime helper should clamp batch floor and normalize invalid duration/end values."""
         host = _Host()
--- a/acestep/core/generation/handler/generate_music_request_test.py
+++ b/acestep/core/generation/handler/generate_music_request_test.py
@@ -47,21 +47,25 @@ class GenerateMusicRequestMixinTests(unittest.TestCase):
         self.assertEqual(task, "cover")
         self.assertNotEqual(instruction, "old")
 
-    def test_neutralize_cover_params_reset_for_non_cover_tasks(self):
-        """Non-cover tasks must drop leaked cover params to neutral (issue #1271)."""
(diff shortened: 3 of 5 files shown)


PythonToGo:
Thanks for the review 

On the guard scope — this is intentional. _neutralize_cover_only_params currently resets audio_cover_strength / cover_noise_strength only for text2music, and preserves them for every other task (cover, cover-nofsq, repaint, lego, extract, complete).

Rationale: text2music is the only task whose source latent is silence, so a non-zero cover_noise_strength leaking in from a prior Remix/Cover session has no real content to anchor to and degrades into full-duration noise; exactly the #1271 failure. The other tasks all operate on real source audio/codes, where these knobs may be meaningful controls rather than stale leftovers, so I deliberately kept the blast radius to the one task that actually reproduces the bug.

I'm aware the issue note suggested neutralizing "unless the resolved task is cover/cover-nofsq" (i.e. also for repaint/lego/extract). I went narrower on purpose to avoid changing behavior on the source-audio tasks. If you'd prefer the broader guard, I'm happy to switch, so just say the word and I'll update it.

ChuxiJ (changes requested):
Thanks for tightening the scope to `text2music`; that direction makes sense for #1271.

I cannot merge this yet because the focused regression suite exposes a real integration/test-contract break. In a clean temporary worktree on this PR, running:

`uv run --no-sync python -m unittest acestep.core.generation.handler.generate_music_request_test acestep.core.generation.handler.generate_music_test`

fails with 5 errors in `generate_music_test`: `GenerateMusicMixin.generate_music()` now calls `self._neutralize_cover_only_params(...)`, but the existing `_Host` test double for `GenerateMusicMixin` does not provide that method.

Please update the mixin/test composition so this helper is available wherever `GenerateMusicMixin` is exercised, or adjust the helper placement so `GenerateMusicMixin` does not depend on a method outside its tested contract. The CodeRabbit comment wording nit is minor, but this test failure should be fixed before merge.

PythonToGo:
> Thanks for tightening the scope to `text2music`; that direction makes sense for #1271.
> 
> I cannot merge this yet because the focused regression suite exposes a real integration/test-contract break. In a clean temporary worktree on this PR, running:
> 
> `uv run --no-sync python -m unittest acestep.core.generation.handler.generate_music_request_test acestep.core.generation.handler.generate_music_test`
> 
> fails with 5 errors in `generate_music_test`: `GenerateMusicMixin.generate_music()` now calls `self._neutralize_cover_only_params(...)`, but the existing `_Host` test double for `GenerateMusicMixin` does not provide that method.
> 
> Please update the mixin/test composition so this helper is available wherever `GenerateMusicMixin` is exercised, or adjust the helper placement so `GenerateMusicMixin` does not depend on a method outside its tested contract. The CodeRabbit comment wording nit is minor, but this test failure should be fixed before merge.

Thanks for catching this and running the focused suite; confirmed and fixed :)

The `_Host` test double in `generate_music_test.py` predates this PR and didn't compose `GenerateMusicRequestMixin`, so it had no `_neutralize_cover_only_params` for `generate_music()` to call. Added a stub mirroring the real text2music-only contract (same per-method stub pattern the file already uses). `generate_music_dcw_test.py` imports the same `_Host`, so its errors were the same root cause and are fixed too. Also applied the comment-wording nit from CodeRabbit.

`uv run --no-sync python -m unittest acestep.core.generation.handler.generate_music_request_test acestep.core.generation.handler.generate_music_test` now passes clean (22/22).

PTAL 🙏🏻 :)

ChuxiJ (approved):
Thanks for the quick follow-up. I re-ran the focused regression suite on the updated PR head `9cb245f`:

```bash
python -m unittest \
  acestep.core.generation.handler.generate_music_request_test \
  acestep.core.generation.handler.generate_music_test
```

Result: 22 tests ran and passed.

The earlier mixin/test-contract blocker is fixed, and the text2music-only guard is the right narrow scope for #1271. Approving.

PR #1307, opened 2026-08-28
fix: _get_project_root() resolves two directories too shallow

zu6fx:
### Summary
The randomize button (and /create_random_sample directly) always
fails with "No example data available". Turned out _get_project_root()
in api_routes.py only walks up 3 directories from api_routes.py's own
path, which puts you at acestep/ui, not the repo root. So it looks for
examples/simple_mode two levels short of where it actually lives, finds
nothing, and the example caches load empty at startup.

### Scope
One line, api_routes.py only. Nothing else touched.

### Risk and Compatibility
Just a dirname() count, no platform-specific logic involved. Doesn't
touch anything outside this one function.

### Regression Checks
- Confirmed the corrected path lands on the real repo root and
  examples/simple_mode is found there.
- File still compiles.
- /create_random_sample returns actual example data instead of erroring.

### Reviewer Notes
Nothing else needed here, self-contained fix.

## Summary by CodeRabbit

* **Refactor**
  * Improved the application’s project path resolution for clearer, more maintainable behavior.
  * No user-facing functionality or path resolution behavior has changed.
* **Documentation**
  * Expanded technical documentation for project root resolution.

### DIFF
--- a/acestep/ui/gradio/api/api_routes.py
+++ b/acestep/ui/gradio/api/api_routes.py
@@ -127,8 +127,14 @@ atexit.register(_close_result_cache)
 # =============================================================================
 
 def _get_project_root() -> str:
-    """Get project root directory"""
-    return os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))))
+    """Get project root directory.
+
+    Returns:
+        Absolute path to the repository root, five directories up from
+        this file (acestep/ui/gradio/api/api_routes.py).
+    """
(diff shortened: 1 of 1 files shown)


ChuxiJ (approved):
Thanks, this is a clean narrow fix.

I verified the updated project-root helper in a clean worktree on PR head `0de25ff`:

```bash
python - <<'PY'
from acestep.ui.gradio.api.api_routes import _get_project_root, _load_all_examples
# root resolves to the repository root and simple_mode examples load
PY
```

The helper resolves to the repository root and `_load_all_examples("simple_mode")` loads 200 examples. Approving.

PR #1308, opened 2026-08-29
fix(ui): Fix Remix no_fsq control layout and tooltip clipping

dd178:
## Problem
In the WebUI Remix mode, the `no_fsq` parameter renders incorrectly: it is wrapped in a very short scroll strip and its hover tooltip cannot display normally (it hides inside the scroll strip and requires scrolling to see). The root cause is that the Remix-only controls (help tutorial, mix strength, cover strength, no_fsq) are scattered across separate containers, and `gr.Group` / `.styler` are `overflow: hidden` by default, clipping the absolutely-positioned tooltip into a thin strip.

## Change
- Group the Remix-only controls (help tutorial, mix strength, cover strength, no_fsq) into a single `gr.Group`, laid out like the Custom tutorial: help `?` at the top, then 混音强度 / 翻唱强度 / no_fsq stacked below. These are Remix-only params now shown only in Remix mode.
- Add `.gr-group` and `.styler` to the `:has(.has-info-container)` overflow:visible rule so the no_fsq info tooltip is not clipped by the group's overflow-hidden boundary.

## Verification
- `acestep/ui/gradio/` contract tests pass. This is a pure UI layout/CSS change with no business logic, so no new assertion is strictly required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tooltip visibility by preventing clipping across additional interface elements.
  * Improved the layout of remix controls so help information displays correctly alongside related settings.

### DIFF
--- a/acestep/ui/gradio/interfaces/__init__.py
+++ b/acestep/ui/gradio/interfaces/__init__.py
@@ -154,7 +154,9 @@ def create_gradio_interface(dit_handler, llm_handler, dataset_handler, init_para
         .accordion:has(.has-info-container),
         .tabs:has(.has-info-container),
         .gr-block:has(.has-info-container),
-        .gr-box:has(.has-info-container) {
+        .gr-box:has(.has-info-container),
+        .gr-group:has(.has-info-container),
+        .styler:has(.has-info-container) {
             overflow: visible !important;
             contain: none !important;
         }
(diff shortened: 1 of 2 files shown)


ChuxiJ (approved):
Thanks, this is a focused UI layout fix.

I ran the relevant Gradio interface/mode tests on PR head `4d93422`. The narrowed suite has one existing failure in `generation_decomposition_contract_test`, but I reproduced the same failure on clean `origin/main`, so it is not introduced by this PR.

The diff is limited to grouping Remix-only controls and expanding the tooltip overflow selectors. CodeRabbit has no actionable comments on the current head. Approving.

PR #1309, opened 2026-08-29
fix(flow-edit): Fix device mismatch when DiT offload meets flow-edit morph

dd178:
## Problem
With "Offload DiT to CPU" enabled, the `flow_edit_morph=True` path calls the downstream `prepare_condition` (used for auto-LRC + scoring) outside the `_load_model_context("model")` block. At that point the whole model — including the condition encoder's `text_projector` — has already been offloaded back to CPU, while `text_hidden_states` / `silence_latent` remain on cuda:0, so `F.linear` raises a device mismatch. This only triggers on the `flow_edit_morph` path.

## Change
- Wrap the downstream `prepare_condition` (and `attn` construction) in a `_load_model_context("model")` block so the condition encoder and `silence_latent` are back on the device for that call.
- Add a regression test: `ModelContextTrackingHandler` tracks the `_load_model_context("model")` activation depth and asserts the downstream `prepare_condition` runs while the model is loaded.

## Verification
- `service_generate_flow_edit_test.py` passes, including the new offload regression test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved flow-edit audio generation when model offloading is enabled.
  * Ensured generated conditions are prepared correctly for text-to-music and cover workflows.

* **Tests**
  * Added coverage verifying that model resources are available during condition preparation.
  * Expanded task-selection coverage for supported cover, complete, and LEGO workflows.

### DIFF
--- a/acestep/core/generation/handler/service_generate_flow_edit.py
+++ b/acestep/core/generation/handler/service_generate_flow_edit.py
@@ -165,20 +165,25 @@ def dispatch_flow_edit_overlay(
             )
 
     # Return target-side encoder/context for downstream auto-LRC + scoring.
-    attn = torch.ones(bsz, seq, device=device, dtype=dtype)
-    enc_hs, enc_am, ctx = handler.model.prepare_condition(
-        text_hidden_states=payload["text_hidden_states"],
-        text_attention_mask=payload["text_attention_mask"],
-        lyric_hidden_states=payload["lyric_hidden_states"],
-        lyric_attention_mask=payload["lyric_attention_mask"],
-        refer_audio_acoustic_hidden_states_packed=payload["refer_audio_acoustic_hidden_states_packed"],
-        refer_audio_order_mask=payload["refer_audio_order_mask"],
-        hidden_states=ctx_input,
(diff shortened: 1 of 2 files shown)


ChuxiJ (changes requested):
Thanks for the focused flow-edit/offload fix. The runtime direction makes sense, but I cannot merge this yet because the focused test suite is failing on the current PR head `dfb49e3`.

Verification run:

```bash
python -m unittest acestep.core.generation.handler.service_generate_flow_edit_test
```

Result: 10 tests ran, 1 failure.

The failure is `CoverStillRunsLmTests.test_no_edit_in_skip_lm_tasks`: it still expects the old literal `skip_lm_tasks = {"cover", "cover-nofsq", "repaint", "extract"}` in `inference.py`, but current main uses `DIRECT_CONDITIONING_TASKS` and includes `complete` / `lego` after the direct-conditioning fix.

Please update that test to assert the current contract rather than the stale source string, and also address the small docstring/lint notes from the existing review. Once the focused suite is green, this should be straightforward to re-review.

PR #1310, opened 2026-08-29
fix(gpu_config): Fix KV cache inflation on service reinit by sizing ratio from live tensors

dd178:
## Problem
`get_lm_gpu_memory_ratio` computed `current_usage = total - free` (driver view), which counts the CUDA caching allocator's inactive pool plus the fixed CUDA-context floor. Both grow across reloads, inflating the ratio → a larger KV cache on each re-init (bounded but convergent) and less headroom for DiT inference. Measured: ratio 0.635 → 0.687, KV cache 2.95 → 3.77 GB on reload, post_kv_free 4.45 → 3.30 GB.

## Change
- `current_usage` is now derived from `torch.cuda.memory_allocated()` (live tensors), the same view `allocate_kv_cache` already uses (`model_runner.py`).
- The DiT/VAE/text live footprint is byte-identical across reloads (+0.00 GB measured), so the ratio and KV cache stay stable on re-init; each LM size returns to its design budget (`kv_cache_4k + 0.3`).
- Add a regression test asserting the ratio derives from live tensors and is invariant to total-free (fragmentation/context) drift.

## Verification
- `gpu_config_lm_memory_ratio_test.py` passes (3 assertions: live-allocated source, free-drift invariance, uncapped tracking).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU memory usage calculations for language model capacity reporting.
  * Excludes cached inactive memory and fixed CUDA overhead from reported current usage, providing more accurate ratios.
  * Preserves safety behavior when available GPU memory is limited.

* **Tests**
  * Added coverage for accurate GPU memory ratio calculations across capped and uncapped memory scenarios.

### DIFF
--- a/acestep/gpu_config.py
+++ b/acestep/gpu_config.py
@@ -1041,7 +1041,14 @@ def get_lm_gpu_memory_ratio(
             # Convert to ratio of total GPU memory
             # nano-vllm uses: target_total_usage = total * gpu_memory_utilization
             # We want: (total * ratio) = current_usage + usable_for_lm
-            current_usage_gb = actual_total_gb - free_gb
+            # current_usage = live tensors in the process (memory_allocated), i.e. the
+            # DiT/VAE/text baseline at LM-init since unload() tears down the prior LM
+            # engine first. NOT the driver's total-free, which counts the caching
+            # allocator's `inactive` pool plus the fixed CUDA-context floor — both
+            # grow across reloads and inflate the ratio -> KV cache. The live-tensor
+            # view matches allocate_kv_cache (model_runner.py), so re-init no longer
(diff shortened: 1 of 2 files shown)


ChuxiJ (approved):
Thanks, this is a much narrower and easier-to-validate version of the LM memory-ratio fix.

I verified PR head `604f99a` with the focused regression suite:

```bash
python -m unittest acestep.gpu_config_lm_memory_ratio_test
```

Result: 3 tests ran and passed.

The change is localized to `get_lm_gpu_memory_ratio`, and using live allocated CUDA memory matches the intended reinit/KV-cache stability contract. The existing `gpu_config.py` size is still a broader cleanup item, but I do not think it should block this targeted fix. Approving.

PR #1311, opened 2026-08-29
fix(llm): Fix LM engine memory leak on web UI service reinit

dd178:
## Problem
On re-init the nano-vLLM engine was never truly released: `LLMHandler.unload()` only `reset()` the block allocator (keeping the engine and its KV cache / CUDA graphs alive) instead of calling the real `LLMEngine.exit()` teardown, and `atexit` retained the old engine. Each reload therefore leaked the previous engine's weights, KV cache, and CUDA graphs; the KV cache re-allocation then shrank as free VRAM dropped, producing the reported convergent memory growth.

## Change
- `llm_inference.py`: for the vLLM backend, `unload()` now calls `self.llm.exit()` (full engine teardown) instead of `reset()`, then deregisters the engine's atexit callback so process shutdown does not re-run `exit()` on a released runner (which would log a noisy AttributeError). `initialize()` calls `unload()` first so a reinit releases the old engine before building a new one.
- `lm_score.py`: the PMI scoring temporary-unload helper now uses `llm_handler.unload()` instead of a manual reset + reference drop.
- `service_init.py`: when `init_llm` is disabled, unload a previously-loaded LM so it does not stay resident across re-initialization.
- `lm_score_test.py`: update the PMI offload test to assert `unload()` is called.
- Add `llm_inference_unload_test.py` (5 regression tests): assert the vLLM engine is `exit()`-ed (not `reset()`-ed), the atexit callback is deregistered, teardown errors are swallowed, and non-vLLM backends are unaffected.

## Verification
- `lm_score_test.py` + `llm_inference_unload_test.py` all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved language model resource cleanup by fully releasing model memory, caches, and GPU execution resources.
  * Prevented previously loaded models from remaining active when language model initialization is disabled or restarted.
  * Added safer handling for cleanup failures, preventing a new model from starting when old resources cannot be released successfully.
  * Improved recovery and state consistency after model shutdown errors.

### DIFF
--- a/acestep/core/scoring/lm_score.py
+++ b/acestep/core/scoring/lm_score.py
@@ -168,22 +168,13 @@ def _temporary_unload_interactive_lm_for_scoring(llm_handler):
         return
 
     logger.info("[scoring] Temporarily unloading vLLM runtime for PMI scoring")
-    llm_runtime = llm_handler.llm
+    # unload() now performs a full engine teardown (exit()), releasing the model
+    # weights, KV cache, and CUDA graphs so the HF scorer has room (and so the
+    # subsequent initialize() restore does not leave a stale engine behind).
     try:
-        if hasattr(llm_runtime, "reset"):
-            llm_runtime.reset()
+        llm_handler.unload()
     except Exception as exc:
-        logger.warning("[scoring] vLLM reset during PMI offload failed: {}", exc)
-    try:
-        llm_handler._cleanup_torch_distributed_state()
-    except Exception as exc:
-        logger.warning("[scoring] vLLM distributed cleanup during PMI offload failed: {}", exc)
-    llm_handler.llm = None
-    llm_handler.llm_initialized = False
-    gc.collect()
-    if torch.cuda.is_available():
-        torch.cuda.empty_cache()
-        torch.cuda.synchronize()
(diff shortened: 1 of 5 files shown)


ChuxiJ (changes requested):
Thanks, this is the right area to fix and the focused tests pass locally on PR head `9aee9f1`:

```bash
python -m unittest \
  acestep.llm_inference_unload_test \
  acestep.core.scoring.lm_score_test
```

Result: 10 tests ran and passed.

However, I still cannot merge this because the teardown-failure path is unsafe for exactly the lifecycle problem this PR is trying to fix. If `self.llm.exit()` raises, `unload()` currently logs the failure but still clears `self.llm`, `llm_backend`, and related state. Then `initialize()` can continue and create a replacement engine while the previous vLLM runtime may not have fully released weights, KV cache, CUDA graphs, workers, or distributed state.

Please make that failure state explicit before merge. Either:

- preserve the existing engine/handler state when `exit()` fails so a retry can act on the same runtime; or
- have `unload()` return a failure signal/raise a controlled exception, and make `initialize()` stop instead of building a new engine after incomplete teardown.

Please add a regression test for `initialize()` after failed teardown as well. Once that edge is closed, this should be a strong high-priority fix.

PR #1313, opened 2026-08-29
Separate generation success from export-format failure

atlas-ocm:
A failed MP3 export (e.g. missing ffmpeg) currently discards an already-successfully-generated audio result and surfaces as a hard error in the Gradio UI, indistinguishable from a real generation failure (GPU OOM, NaN latents, etc). The DiT/VAE pipeline already completed successfully in this case; only the WAV -> MP3 format conversion step failed.

`_save_mp3()` already writes a temporary WAV file before invoking ffmpeg. This PR preserves that WAV instead of deleting it when ffmpeg fails (missing binary, timeout, or non-zero exit), and introduces `AudioExportDegradedError` carrying the fallback WAV path plus the original failure reason.

`generate_with_progress()` catches this specific exception, uses the WAV fallback as the sample's saved path, and reports a distinguishable status (`Generation Complete (WAV). <key>: mp3 export unavailable (<reason>), saved as WAV`) instead of letting the exception propagate into a generic error toast that discards the result.

## Behavior

* generation PASS + export PASS -> unchanged: `Generation Complete`, requested format saved
* generation PASS + export FAIL -> `Generation Complete (WAV). <reason>`, WAV saved instead of discarding a valid result
* generation FAIL -> unchanged: existing error handling

Other export formats (opus/aac/flac/wav) and the case where the WAV fallback move itself fails are unaffected -- they keep their existing behavior (soundfile fallback / hard failure respectively).

## Verification

Reproduced live on a GTX 1080 Ti deployment:

* ffmpeg installed: MP3 export produces a valid file (ffprobe-verified, 15.0s duration)
* ffmpeg simulated missing: WAV fallback is written and playable (192KB, valid audio) instead of the result being discarded

Tests cover: missing-ffmpeg WAV preservation, timeout WAV preservation, `save_audio()` propagating the degraded error with a valid WAV path, and the case where the fallback move itself fails (falls back to the original hard-failure behavior with the real ffmpeg reason surfaced).

## Summary by CodeRabbit

* **New Features**
  * Preserves a playable WAV fallback when requested audio export formats, such as MP3, cannot be generated.
  * Displays a completion status with export warnings when fallback audio is used.
  * Supports explicit CUDA precision settings through `ACESTEP_DTYPE`.

* **Bug Fixes**
  * Prevents failed audio exports from discarding successfully generated samples.
  * Provides clearer errors for invalid or unsupported CUDA precision settings.

### DIFF
--- a/acestep/audio_utils.py
+++ b/acestep/audio_utils.py
@@ -11,6 +11,7 @@ Independent audio file operations outside of handler, supporting:
 import io
 import json
 import os
+import shutil
 import subprocess
 import hashlib
 import tempfile
@@ -20,6 +21,21 @@ import torch
 import numpy as np
 import torchaudio
 from loguru import logger
+
+
+class AudioExportDegradedError(RuntimeError):
+    """Raised when audio generation succeeded but the requested export format failed.
+
+    The underlying audio was already synthesized successfully; only the
+    format conversion (e.g. WAV -> MP3 via ffmpeg) failed. ``wav_fallback_path``
+    points to a valid WAV file already saved to disk, so callers can present
+    this as a degraded-but-successful result instead of a hard failure.
+    """
+
+    def __init__(self, message: str, wav_fallback_path: str, requested_format: str):
+        super().__init__(message)
+        self.wav_fallback_path = wav_fallback_path
+        self.requested_format = requested_format
 
 
 def apply_fade(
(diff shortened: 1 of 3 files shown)


atlas-ocm:
Note: this PR is stacked on top of #1312 (`fix/cuda-dtype-env-override`) — the diff currently includes both commits since this branch was cut from `main` after that commit landed locally. Once #1312 merges, this diff will narrow to just the export-status changes (`audio_utils.py`, `audio_utils_test.py`, `generation_progress.py`).

